PR #30, opened 2017-05-31
refactor(utils): Simplify align/justify

obartra:
Simplify `getJustify` and `getAlignment` methods


src/utils.js
```diff
@@ -17,24 +17,26 @@ export function times(number = 0, callback = noop) {
...
-      return `${prefix}bottom`
+      return `${prefix}--bottom`
     default:
       return ''
   }
 }
 
-export function getJustify(value, prefix) {
+export function getJustify(value) {
+  const prefix = 'grid'
```
arahansen: might as well move this to the top of the file

obartra: hm this is specific to justify not to utils though

arahansen: But creating new variables  on every call Oscar! Every call!!

arahansen: But no, you're right. I concede 😄

obartra: haha well if we are going to get into optimizations, external variables will de-opt the function from optimization!!

PR #33, opened 2017-06-02
fix(tests): Add babel-polyfill

arahansen:
jest does not run with any polyfills. We need them due to Object.values


arahansen:
actually I'm silly. I was running an old version of node.

Issue #2, opened 2017-05-28
Add page layout components


obartra:
- [x] Page layout components missing (`gl-grid-root` / `gl-grid-content` should not be necessary).
- [x] Page layout stories missing
- [x] Page layout needs to be reworked
- [x] Allow for header / body / footer, holy grail layout...,

PR #34, opened 2017-06-02
feat(layout): Add basic layouts

### LINKED ISSUE #15: Gridlex / Bulma grid
- [ ] Bulma layout system seems to cover most of our needs and it's simpler. Consider some of their approaches
- [x] Gridlex grid system also seems simpler, evaluate

http://bulma.io/documentation/grid/columns/
https://github.com/jgthms/bulma/blob/master/sass/grid/columns.sass

obartra:
Add vertical layout system

Current implementation example looks like this:

http://recordit.co/ipNR6T0Bzb

- add basic vertical layout system + usage example
- limit the width of its children to max page width and center
- update examples to use this layout
- add Holy Grail layout example

Closes #15, #2


.vscode/settings.json
```diff
@@ -0,0 +1,10 @@
+{
```
arahansen: could we add prettier settings in here too?

package.json
```diff
@@ -13,8 +13,7 @@
     "babel-plugin-transform-class-properties": "6.24.1",
     "babel-plugin-transform-object-rest-spread": "6.23.0",
     "commitizen": "2.9.6",
-    "css-loader": "0.28.4",
```
arahansen: we will need css-loader if we want css-modules

src/grid.hoc.js
```diff
@@ -1,26 +1,21 @@
...
+import { compact, getDisplayName, getJustify, getAlignment } from './utils'
+import { type Offset, type Size } from './types'
 import { ALIGN, JUSTIFY } from './values'
 
 export default function Grid(Component: any) {
   return class withGrid extends React.PureComponent {
     props: {
       align?: Symbol,
-      bottom?: boolean,
+      bottom: boolean,
```
arahansen: removing the `?` means the prop is required. I thought bottom, stretch, and margin are optional.

src/layout.hoc.js
```diff
@@ -0,0 +1,44 @@
+// @flow
+import React from 'react'
+import { compact, getDisplayName } from './utils'
+import { SIZE } from './values'
+
+function getSize(size) {
```
arahansen: Missing types, though not sure if we want to set a linting rule to require types. Flow might have good enough type inference here to know that the param has to be a Symbol for example.

stories/grid/autoflow.js
```diff
@@ -18,27 +18,22 @@ story.add('Auto Flow', () => {
...
-              }}
-            />
-            {times(
-              number('items', 5, { range: true, min: 0, max: 100 }),
-              index => <Box size={2} key={index} type={getBoxType(index)} />
-            )}
-          </Grid>
-        </div>
-      </div>
+      <Root stretch={stretch} margin={margin} bottom={bottom}>
```
arahansen: Cool thing to try here `<Root {...{ stretch, margin, bottom }} />`

stories/grid/align.js
```diff
@@ -37,56 +38,52 @@ story.add('Vertical Align', () => {
...
-            <Box size={2} type="C" value="MIDDLE" align={Item.ALIGN.MIDDLE} />
-            <Box size={2} type="D" value="BOTTOM" align={Item.ALIGN.BOTTOM} />
-            <Box size={2} type="C" value="MIDDLE" align={Item.ALIGN.MIDDLE} />
-            <Box size={2} type="E" value="DEFAULT" />
-            <ReferenceColumn height={height} />
+      <Root>
+        <h1>Item Align</h1>
+        <Grid margin={margin} stretch={stretch}>
+          <ReferenceColumn height={height} />
+          <Box size={2} type="B" value="TOP" align={Item.ALIGN.TOP} />
```
arahansen: Shouldnt these Symbols be available on Box?

.vscode/settings.json
obartra: not sure how, does vsc handle it for us?

package.json
obartra: good catch, will update

.vscode/settings.json
arahansen: I have vs code settings for noSemi, spaces, and single quotes.

src/grid.hoc.js
obartra: Oh I thought because default props set it it would never be not set. is that a correct assumption or does this just rely on statically checking if the values are passed?

src/layout.hoc.js
obartra: good point, is there anyway we can get some linting errors when types are missing?

stories/grid/autoflow.js
obartra: this is some sexy destructuring, will update

stories/grid/align.js
obartra: sure yeah we can do that too, I wanted to be explicit that those came from Item and not something else defined by Box but I can see a case to do it based of Box too since I'm using Box here

obartra:
Regarding nesting, it needs some love (https://github.com/obartra/flexgrid/issues/31) but in general all grid systems support some way or another of nesting. I agree we should add more examples that showcase how to use it properly

.vscode/settings.json
obartra: that's awesome, do you need a prettier-vscode plugin for it?

src/grid.hoc.js
arahansen: I believe this refers specifically to props being passed in.

src/layout.hoc.js
arahansen: yeah eslint has a suite of flow based rules we will want to investigate

obartra: makes sense, will update

src/grid.hoc.js
obartra: makes sense, will update

.vscode/settings.json
obartra: added

obartra:
Feedback addressed on [this commit](https://github.com/obartra/flexgrid/pull/34/commits/db170500f7c19ccdd7f2b3c97cef3b99f89f1814) rebasing

PR #35, opened 2017-06-04
refactor(flow): add type checking via flow

arahansen:
We don't necessarily need to merge this. I mostly was trying this out as an experiment in how flow would integrate into the existing workflow.

It ends up being pretty nice in my opinion.

- More terse, and readable prop type definitions and doesn't rely on external package
- Better intellisense including knowledge of whats being imported from other modules
- Superb type inference even before you explicitly provide types
- Inline/"compile time" errors for invalid types via VS Code plugin and doesn't rely on runtime like prop-types
- Fully opt-in via `// @flow` so if for some reason providing types for a file doesn't make sense, it's simple to turn off type checks
- Even in a smaller project such as this, it managed to catch a potential [bug](https://github.com/obartra/flexgrid/compare/master...arahansen:flow?expand=1#diff-c250b34c0c7cf11397c8fc3261491b29L41)!

To be fair, there is a learning curve to understanding the syntax. And there are some gotchas (example, props + defaultProps type checking seem somewhat cumbersome -- but this could be considered part of the learning curve).

While the documentation is great, if you need to do anything that perhaps fall out of their documentation, help seems a bit hard to find. I wish they had a gitter or other chat channel.

It also took a bit of fiddling to get thing set up correctly with my editor + configuring flexgrid for flowtypes.

So far, I have found that the gains in developer experience, and additional guarantees we can make about our codebase outweigh potential downsides. Also being given escape hatches (turning off types, or defaulting to `any` in extreme cases) should help keep things moving along.

Like any new technology - especially complex, sophisticated systems like flow - it takes time to understand how it all fits together. But I see real potential here for a great developer experience.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,8 +1,8 @@
 {
   "extends": [
-      "airbnb",
-      "prettier",
-      "prettier/react"
+    "airbnb",
+    "prettier",
+    "prettier/react"
   ],
   "parserOptions": {
     "ecmaVersion": 2016,
@@ -16,7 +16,8 @@
     "browser": true
   },
   "plugins": [
-    "react"
+    "react",
+    "flowtype"
   ],
   "settings": {
     "react": {
@@ -26,6 +27,7 @@
   "parser": "babel-eslint",
   "rules": {
     "react/jsx-filename-extension": 0,
-    "import/no-extraneous-dependencies": 0
+    "import/no-extraneous-dependencies": 0,
+    "flowtype/define-flow-type": 2
   }
 }
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,0 +1,8 @@
+[ignore]
+.*/node_modules/stylelint/.*
+
+[include]
+
+[libs]
+
+[options]
--- a/package.json
+++ b/package.json
@@ -18,9 +18,11 @@
     "eslint": "3.19.0",
     "eslint-config-airbnb": "15.0.1",
     "eslint-config-prettier": "2.1.1",
+    "eslint-plugin-flowtype": "^2.34.0",
     "eslint-plugin-import": "2.3.0",
     "eslint-plugin-jsx-a11y": "5.0.3",
     "eslint-plugin-react": "7.0.1",
+    "flow-bin": "0.47.0",
     "husky": "0.13.4",
     "jest": "20.0.4",
     "lint-staged": "3.5.1",
@@ -64,13 +66,13 @@
     "lint": "npm-run-all --parallel lint:*",
     "precommit": "lint-staged",
     "start": "start-storybook -p 9001 -c storybook",
-    "test": "jest"
+    "test": "jest",
+    "flow": "flow"
   },
   "dependencies": {
     "@storybook/addon-knobs": "3.0.0",
     "@storybook/addon-notes": "3.0.0",
     "@storybook/react": "3.0.0",
-    "prop-types": "15.5.10",
     "react": "15.5.4",
     "react-dom": "15.5.4"
   },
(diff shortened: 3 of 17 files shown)


obartra:
I like it! And good catch on the bug!

- I've a couple questions, could we turn it on by default?
- Is the `flow` task meant as built step? If so, should it be part of a webpack build?
- Do both tasks (`start` and `build` work with it?)

arahansen:
These are all great questions. I'll answer them the best I can in order:
**Can we turn flow on by default (instead of having to add anything to the file)?**
I had looked into this possibility, but didn't find very many good answers. I think we have two options (kinda):
  - We can include an eslint rule that has an autofixable option to add `// @flow` to the beginning of all files. So running `eslint --fix` will add the line to all files that don't have it
  - there is an `flow check --all` but I'm not sure this is a feasible option. My computer just hangs when trying to do this as I think it tried checking node_modules regardless of `[ignore]` options

Neither of these solutions are exactly what we want, but are a decent enough band-aid perhaps. I'm open to other explorations.

**Is the flow task meant as built step? If so, should it be part of a webpack build? Should it run on CI?**
I think it would make sense to include it as a build step. There seems to be a couple good options here:
- [flowtype-loader](https://github.com/torifat/flowtype-loader)
- [tcomb-babel-plugin](https://github.com/gcanti/babel-plugin-tcomb)

I haven't tried these, and we may be unhappy with how much time it adds to our watch step. So may consider one of the following:
# Only run loader for CI build and rely on IDE feedback during development
# Run loader in parallel and get feedback in terminal, but dont block bundling

**Do both tasks (start and build work with it?)**
As this PR stands, start and build both work since the types are just ignored and stripped out. Ideally, CI would run type checking and fail if any errors are found. 

**Is there a Component type? If not, can we define a type for "function / Component extended class" as we would do on TS?**
There are a couple ways to define Components (might be good to set a standard there). In this PR I added a `prop` attribute to the class you can also do the same for `state` typing.

The other approach is similar to TypeScript where you define your types as `... extends Component<DefaultProps, Props, State>` where each definition is a type declared independently.

I ran into issues defining Component types for HOCs. I didn't find much in documentation on how best to declare these types. We may need to define our own "Component as parameter" types that flow will be happy with. I actually went with `any` here since I found an open source Facebook app that had a similar situation. [See here](https://github.com/fbsamples/f8app/blob/3e5a0c88cd83cfabfca556c26138b1ce2537b4f2/js/common/ItemsWithSeparator.js#L35). There are likely more examples in there worth studying!

**Can/Should we get rid off any any (either by setting the type or removing the type)?**
`any` should definitely be used sparingly. Flow also has a `mixed` type which may be more appropriate in many circumstances. (For instance, one spot I said `any[]` when `mixed[]` was probably more accurate). `any` can be a nice escape hatch, and may be overly limiting to completely ban it.

obartra:
**Can we turn flow on by default (instead of having to add anything to the file)?**
- Would something like `git ls-files *.js | xargs flow check-contents` work to parse individual files? If that works we may want to also run it as a pre push check
- Good point, we should run `eslint --fix` as precommit hook anyway, that would make sure that when it runs on CI, if there are any issues, things fail rather than skipping the file

**Is the flow task meant as built step? If so, should it be part of a webpack build? Should it run on CI?**
- Got it, yeah I agree it makes sense to combine it with the `lint` task locally. That would also ensure it runs on CI too (since it checks for linting issues already)
- If code is already stripped out automatically then yeah I agree no need to add additional loaders. Particularly when the editor should provide the same feedback already

**Is there a Component type? If not, can we define a type for "function / Component extended class" as we would do on TS?**
- Makes sense, we also don't need to fix every single issue on this PR so as you have it now looks good to me

**Can/Should we get rid off any any (either by setting the type or removing the type)?**
- Same thing, we can take another pass after this PR lands

PR #37, opened 2017-06-05
refactor(css): update css to use css-modules

arahansen:
- flex.css and storybook styles now use css-modules.
- Update jest module mapper to use `identity-obj-proxy` for mocking css-module classNames.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "eslint-plugin-react": "7.0.1",
     "flow-bin": "0.47.0",
     "husky": "0.13.4",
+    "identity-obj-proxy": "^3.0.0",
     "jest": "20.0.4",
     "lint-staged": "3.6.0",
     "npm-run-all": "4.0.2",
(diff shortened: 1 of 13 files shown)


storybook/webpack.config.js
```diff
@@ -5,7 +5,11 @@ module.exports = {
     rules: [
       {
         test: /\.css$/,
-        loaders: ['style-loader', 'css-loader', 'postcss-loader'],
+        loaders: [
+          'style-loader',
+          'css-loader?modules&importLoaders=1&localIdentName=[name]__[local]___[hash:base64:5]',
```
obartra: can we use the options syntax instead?

stories/grid/nested.js
```diff
@@ -7,6 +7,7 @@ import Grid from '../../src/grid'
 import { times } from '../../src/utils'
 import Box from '../box'
 import { getBoxType } from '../utils'
+import styles from '../../src/flex.css'
```
obartra: I'm wondering if it's worth adding a linter rule to enforce all named css imports set the variable name to 'styles'

package.json
```diff
@@ -81,4 +81,4 @@
       "path": "cz-conventional-changelog"
     }
   }
-}
+}
```
obartra: I'm adding vsc workspace settings to automatically create new lines here, in the meantime, can you set it to add newlines?

obartra:
Awesome work here!

PR #54, opened 2017-06-07
Obartra/add example

### LINKED ISSUE #47: Offsets are broken
Fix page column offsets. They are broken

obartra:
Code available [here](https://obartra.github.io/flexgrid/branch/obartra/add-example')

- Add more margin and marginSize options for Grid (instead of true/false)
image

- Remove `bottom` parameter
- Add 2 sections page example

image

Closes #47


arahansen:
Maybe we can make Margin more composable?

Allow users to select combos of Top, Bottom, Left, and Right (and maybe All).

And instead of Double for each of them, maybe include a multiplier parameter of sorts?

I'm not liking how many different options there are to keep track of now.

I would rather be able to say "I need Top and Right margin here".

obartra:
In general I want to minimize its usage as much as possible. For almost everything, one should use the default margin. It's only for the edge cases that I want use to have a scape hatch that doesn't feel too hacky.

My rationale for not adding additional parameters (directions, size) is that not all permutations exist for all values. Doing something like `marginSize=[none,half,default,double] margin={top/left/right/bottom/all/default}` doesn't cover all use cases yet the number of permutations grows quickly.

Applying the styling outside CSS would simplify things but is not an option due to CSP concerns.

I'll put some more thought into it and try to come up with something cleaner

src/utils.js
```diff
@@ -35,6 +38,40 @@ export function getAlignment(value: Symbol | void, prefix: string) {
...
+    case MARGIN_SIZE.HALF:
+      return 'Half'
+    case MARGIN_SIZE.DEFAULT:
+    // intentional fall through
+    default:
+      return ''
+  }
+}
+
+export function getMargin(value: Symbol, size: Symbol, prefix: string) {
```
arahansen: Will this function be used for items? Wondering if prefix needs to be passed in.

stories/margin.js
```diff
@@ -0,0 +1,36 @@
...
+  Default: MARGIN.DEFAULT,
+  Horizontal: MARGIN.HORIZONTAL,
+  None: MARGIN.NONE,
+  Vertical: MARGIN.VERTICAL,
+}
+
+export default function getMarginSelect() {
+  const marginStr = select(
+    'Margin',
+    ['Default', 'Horizontal', 'None', 'Vertical'],
```
arahansen: these arrays could just be Object.keys of the maps above

src/utils.js
obartra: I want to use it for individual items as well, I'll file a ticket to keep track of it

stories/margin.js
obartra: oops good call

obartra:
ah but there are tons of margin examples! I added them as knobs on most item examples but we can add specific ones, I'll file a ticket for this as well

arahansen:
oh I see how that works now. That should work 👍

PR #65, opened 2017-06-08
feat(testing): Add image comparison testing

obartra:
Add image comparison testing with screener.io


arahansen:
woah. that was easy.

PR #66, opened 2017-06-11
feat(docs): add usage documentation

arahansen:
- add forked storybook-jsx-addon with better defaultProps support
- add static property to HOCs to allow for accessing wrapped component display name directly

### DIFF
--- a/stories/configStories.js
+++ b/stories/configStories.js
@@ -1,0 +1,31 @@
+// @flow
+import { storiesOf } from '@storybook/react'
+import { withKnobs } from '@storybook/addon-knobs'
+
+function unwrapDisplayNameFromHOC(name: string = '') {
+  return (name.match(/(?:\(([^()]+)\))/) || [])[1] || name
+}
+
+const configStories = (storiesOfName: string, storiesModule: typeof module) => {
+  const stories = storiesOf(storiesOfName, storiesModule)
+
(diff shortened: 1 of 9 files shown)


arahansen:
CI is failing on install. Could this be in part because it's using an older version of yarn?

stories/configStories.js
```diff
@@ -0,0 +1,29 @@
...
+  stories.addDecorator(withKnobs)
+
+  const addStoryWithJSX = (storyName: string, ComponentFn: () => any) =>
+    stories.addWithJSX(storyName, ComponentFn, {
+      showDefaultProps: false,
+      useBooleanShorthandSyntax: true,
+      skip: 1, // assumes each story will have a WithExtensions wrapper, this config option skips that wrapper
+      displayName: element =>
+        /* eslint-disable no-underscore-dangle */
+        element.type.__WRAPPED_COMPONENT_DISPLAY_NAME__ ||
```
obartra: Would it be safe to assume that any parenthesis in the name belong to HOCs and that we only care about the innermost one? If so we could get the name directly by just calling `getDisplayName(element).match(/(?:^[^\(\)]+$|\(([^\(\)]+)\))/)[1]` (using [this regex](https://regex101.com/r/fgJayI/1))

The nice part about this is that it would work without needing any changes to React classes

Issue #36, opened 2017-06-04
Refactor Build Process


arahansen:
flexgrid was originally intended as a pure css solution. Thus has a build system to accommodate such a solution. We should upgrade the build pipeline to take into account the new component-driven nature 

- [ ] Add generalized, production-ready build rather than only Storybook-centric
- [ ] Ensure lodash is code split properly
- [ ] Webpack build should extract JS into standalone min bundle for user consumption by users
- [ ] Add flowtype checks as part as build process
- [ ] Generate flow types and output to dist using something like [flow-copy-source](https://github.com/AgentME/flow-copy-source)

obartra:
Looks good but the webpack build for storybook is needed. We just need an independent one for publication

arahansen:
The two should be sharing many, if not, most config options, correct?

obartra:
Unless they updated it, Storybook uses webpack 1. The current setup also relies on storybook extending the webpack config. I think it's better if we use our own since that one is purely meant as configuration for storybook

PR #67, opened 2017-06-12
refactor(structure): Dynamically require stories

obartra:
## Changes

- Group all src files within index. This should help on https://github.com/obartra/reflex/issues/36
- Add single global variables file
- removed the `stories/[story]/{index,story}.js` files and replaced by `stories/index.js` that imports all
  files from the folder directly
- moved all files within `/stories` that are not part of a story to `/stories/core` and unified the export

## Next steps
- Automatically read any `.txt` file with the file name than the source. If it exists add it as notes
- Automatically wrap all stories with `WithExtensions` without affecting the JSX to string conversion


stories/index.js
```diff
@@ -1,5 +1,65 @@
...
+
+  stories.addDecorator(withKnobs)
+
+  const addStoryWithJSX = (storyName: string, ComponentFn: () => any) =>
+    stories.addWithJSX(storyName, ComponentFn, {
+      showDefaultProps: false,
+      useBooleanShorthandSyntax: true,
+      skip: 1, // assumes each story will have a WithExtensions wrapper, this config option skips that wrapper
+      displayName: element =>
+        unwrapDisplayNameFromHOC(utils.getDisplayName(element.type)),
```
arahansen: the function I had here was slightly different from getDisplay name, which is why I didn't use it originally. This is because we also need to account for primitives (h1, div, p, etc) with `element.type.name`. Although maybe we need to update the util function anyways to account for that.

stories/index.js
```diff
@@ -1,5 +1,65 @@
...
-import './grid'
-import './components'
-import './layout'
+
+/**
+ * Reads the `/components`, `/grid` and `/layout` subfolder files and adds them to their respective
+ * stories
+ */
+
+const folders = {
```
arahansen: could we do a `require.context` on a root folder that will parse all sub directories similar to what we are doing here? That way, we won't have to update this list of folders if we add additional story types.

stories/index.js
obartra: yeah I'll do that

obartra:
I've created https://github.com/obartra/reflex/issues/69 to track moving to lodash, I didn't add it here since it will require build and babel changes to make sure we don't pull the whole library.

src/utils.js
```diff
@@ -7,11 +7,23 @@ const noop = (...params: any[]) => null
...
+export function getDisplayName(
+  WrappedComponent: StringOrWrappedComponent
+): string {
+  const defaultName = 'Component'
+
+  if (typeof WrappedComponent !== 'string') {
+    return WrappedComponent.displayName || WrappedComponent.name || defaultName
+  }
+
+  return WrappedComponent || 'Component'
```
arahansen: Did you mean to use `defaultName` here?

obartra: 😆

obartra:
oops totally misread that, yeah that makes sense. would you define it as `type Component = Class | Function`?

Issue #56, opened 2017-06-08
Header is not sticky


obartra:
Header of the different layouts (Holy Grail, Two Section) is no longer sticky. Fix

obartra:
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Advanced_layouts_with_flexbox

PR #71, opened 2017-06-15
fix(layout): Update layout system

obartra:
These changes simplify the layout to 3 types:

- PARENT: Contains stretch or auto elements, expands to 100% of the container
- STRETCH: Grows to fit, vertical overflow if needed
- AUTO: Size defined by the content, doesn't grow or shrink

To lock a layout element to the top or the bottom of the window you can use: `fixed={Layout.FIXED_POSITION.TOP}` or `fixed={Layout.FIXED_POSITION.TOP}`. This is used by sticky headers while preserving the page scroll.

To change the default overflow setting (overflow: auto) you can set `overflow={Layout.OVERFLOW.NONE}`

Closes https://github.com/obartra/reflex/issues/56

### DIFF
--- a/src/grid.css
+++ b/src/grid.css
@@ -5,10 +5,12 @@
   display: flex;
   flex-flow: row wrap;
   width: 100%;
+  max-width: var(--max-width);
 
   &.gridRoot {
     padding-left: calc(var(--page-margin) - var(--gutter) / 2);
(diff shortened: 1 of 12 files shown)


obartra:
These changes will eventually close https://github.com/obartra/reflex/issues/56

obartra:
I'll go ahead and merge since this fixes a long standing bug but happy to address any concerns in a follow up PR

Issue #72, opened 2017-06-16
consolidate css


obartra:
Layouts should use a single CSS
The color variables are specific to stories, they should not be part of the main css

obartra:
Closed by #74

PR #74, opened 2017-06-16
refactor(design): Update color scheme to match Axons

### LINKED ISSUE #73: use axon colors
The styleguide should use axon branded colors. It should also use Geometria

image

obartra:
Update color theme and font to match Axon styling

image

Closes #73 and https://github.com/obartra/reflex/issues/72


stories/core/designGrid.css
```diff
@@ -55,7 +56,7 @@
     background-image: linear-gradient(
       90deg,
       transparent var(--col),
-      color(var(--blueBerry) alpha(0.1)) var(--gutter)
+      color(var(--bolt) alpha(0.1)) var(--gutter)
```
arahansen: naming convention here should be theme based (ie, `primary1`) to avoid needing to update all references to colors any time the theme colors change.

obartra: I think the color palette is unlikely to change ever on any project unless you are creating something like bootstrap. It's also easy to regexp.

Colors here have no meaning other than showing a bunch of different blocks. I could see naming them something like `color1`, `color2`, `color3` to emphasize their lack of meaning but that's not much different from what we have now and it makes it easier to read the css

Issue #79, opened 2017-06-16
`stretch` should be a property of align for grid


obartra:
Right now Grid takes a boolean parameter to set stretch or not.

This should be part of the ALIGNMENT Symbols and not a separate parameter

obartra:
Closed by https://github.com/obartra/reflex/pull/81

PR #81, opened 2017-06-17
refactor(API): Switch Symbols for Strings

### LINKED ISSUE #78: Remove NONE/DEFAULT Options
These options add additional API surface area while being unclear as to their intent and whether or not they're required.

Instead, we should have default behavior that is implicit through omitting these props.

obartra:
- Add "stretch" as alignment value for Grid (instead of separate boolean parameter)
- Replace all symbols for strings and appropiate flow types
- Set undefined as default except for margin/marginSize where different elements have
  different defaults

Closes #78, #79


storybook/config.js
```diff
@@ -14,3 +14,6 @@ if (typeof window === 'object') {
 
 if (typeof window === 'object')
   window.__screener_storybook__ = require('@storybook/react').getStorybook()
+
+if (typeof window === 'object')
```
obartra: This is a storybook bug. If I'm not careful it keeps automatically adding these 2 lines to this file.

PR #102, opened 2017-06-20
feat(layout): Add report layout

obartra:
- Add padding option to all grids
- Add padding examples
- Add report example

TODO

- Update margin options to use `getSides` for all components
- Polish report example
  - Sticky navigation (https://github.com/obartra/reflex/issues/103)
  - Tabs should not adhere to the grid but to have an option to size to the min size of the content (https://github.com/obartra/reflex/issues/104)
  - Option to increase the separation between items while respecting page margins
Live version available here: https://obartra.github.io/reflex/branch/report (https://github.com/obartra/reflex/issues/105)

image

### DIFF
--- a/src/padding.css
+++ b/src/padding.css
@@ -1,0 +1,22 @@
+@import "./variables.css";
+
+.padding {
+  box-sizing: border-box;
+  width: 100%;
+}
+
+.top {
+  padding-top: var(--gutter);
+}
+
+.right {
+  padding-right: var(--gutter);
+}
+
+.bottom {
+  padding-bottom: var(--gutter);
+}
+
+.left {
+  padding-left: var(--gutter);
+}
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,6 +1,7 @@
 // @flow
+import { uniq } from 'lodash'
 import styles from './index.css'
-import type { Component, Margin, MarginSize } from './types'
+import type { Component, Margin, MarginSize, IndividualSides } from './types'
 
 /* eslint-disable no-unused-vars */
 const noop = (...params: any[]) => null
@@ -49,6 +50,27 @@ export function getMargin(
     default:
       return ''
   }
+}
+
+const individualSides = ['top', 'right', 'bottom', 'left']
+const sideMaps = {
+  horizontal: ['right', 'left'],
+  vertical: ['top', 'bottom'],
+  all: individualSides,
+  none: [],
+}
+
+export function getSides(sides?: string = ''): Array<IndividualSides> {
+  const allSides = sides.split(' ').reduce((acc, current) => {
+    if (individualSides.includes(current)) {
+      return [...acc, current]
+    } else if (current in sideMaps) {
+      return [...acc, ...sideMaps[current]]
+    }
+    return acc
+  }, [])
+
+  return uniq(allSides)
 }
 
 /* eslint-disable no-console */
--- a/stories/grid/padding.js
+++ b/stories/grid/padding.js
@@ -1,0 +1,62 @@
+// @flow
+import React from 'react'
+import { Grid, Item } from '../../src'
+import { Root } from '../core'
+import styles from '../core/stories.css'
+
+export default function() {
+  return (
+    <Root>
+      <Grid justify="center">
+        <Item size={12}>
+          <h1>Top Right Padding</h1>
+        </Item>
+        <Grid size={6} className={styles.colors2} padding="top right">
+          <Item size={12} className={styles.colors3}>
+            Content
+          </Item>
+        </Grid>
+      </Grid>
+      <Grid justify="center">
+        <Item size={12}>
+          <h1>All-sides Padding</h1>
+        </Item>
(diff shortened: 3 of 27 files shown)


stories/layout/report.js
```diff
@@ -0,0 +1,115 @@
...
+  return (
+    <Item size={12} style={{ height }}>
+      <div style={{ height: '100%' }} className={styles.card}>
+        <P>{children}</P>
+      </div>
+    </Item>
+  )
+}
+
+const text = (loremIpsum || '').substr(0, 250)
```
arahansen: why does this need an empty string fallback?

obartra: jest was crashing, but you are right probably the better fix here is figure out why since it should be loading the file stub

stories/layout/stack.js
```diff
@@ -44,7 +44,7 @@ export default function() {
   const sections = number('Sections', 2, { range: true, min: 2, max: 10 })
 
   return (
-    <Layout type="parent" className={style.page}>
+    <Layout type="parent" className={style.page} style={{ maxHeight: '100%' }}>
```
arahansen: I know we don't care about preventing inline styles for devtools like storybook, but is this a pattern we want to allow?

obartra: after discussing with @arahansen, we'll put all css in a file and link it in the notes. That way it's obvious what's needed to generate the code for the page

stories/layout/report.js
```diff
@@ -0,0 +1,111 @@
...
+            <Box size={2} offset={2} type="C" />
+          </Grid>
+        </Layout>
+      </Layout>
+      <Layout type="parent" className={`${styles.main} ${colors2}`}>
+        <Layout type="stretch">
+          <Grid root>
+            <Grid align="stretch">
+              <Grid size={1} className={styles.nav} align="top">
+                <Item size={12} />
```
obartra: This is really hacky, basically adds an element with 0 height so we can use it's margin bottom as margin-top for the next element. We may need an option like `includeMarginTop` but that also sounds really weird

src/utils.js
```diff
@@ -51,6 +52,27 @@ export function getMargin(
...
+    if (individualSides.includes(current)) {
+      return [...acc, current]
+    } else if (current in sideMaps) {
+      return [...acc, ...sideMaps[current]]
+    }
+    return acc
+  }, [])
+
+  return uniq(allSides)
+}
```
obartra: This is the method you suggested @arahansen, now you can use things like: "top horizontal" "all" or "top left right". I think I'll change all margin/paddings to use this on a follow up PR to prevent this one from growing any further

stories/grid/padding.js
```diff
@@ -0,0 +1,62 @@
+// @flow
```
obartra: This is an example to showcase the new padding behavior, looks like this:

image

stories/layout/cards.js
```diff
@@ -25,20 +25,20 @@ export default function() {
...
-            <Box size={2} type="C" />
-            <Box size={2} type="C" />
-            <Box size={2} type="C" />
-            <Box size={2} type="C" />
+            <Box size={2} type="D" />
+            <Box size={2} type="D" />
+            <Box size={2} type="D" />
+            <Box size={2} type="D" />
+            <Box size={2} type="D" />
+            <Box size={2} type="D" />
```
obartra: Changed styling as per @snomu recommendation

stories/layout/layout.css
```diff
@@ -68,3 +76,47 @@
     background-color: var(--bolt);
   }
 }
+
+.report {
```
obartra: this should be in a separate css file

src/padding.css
```diff
@@ -0,0 +1,22 @@
+@import "./variables.css";
+
+.padding {
+  box-sizing: border-box;
+  width: 100%;
+}
+
+.topDirection {
```
arahansen: seems needlessly verbose. maybe just `top`

src/padding.css
```diff
@@ -0,0 +1,22 @@
...
+  margin-top: var(--gutter);
+}
+
+.rightDirection {
+  margin-right: var(--gutter);
```
arahansen: "padding" is an unfortunately in accurate name

src/utils.js
```diff
@@ -1,6 +1,7 @@
 // @flow
```
arahansen: unit tests for utils might be nice at some point

stories/grid/padding.js
arahansen: beautiful

src/padding.css
obartra: yeah it's padding in the sense that affects the same area that would be considered padding in the box model.

src/utils.js
obartra: agreed, this has enough complexity now, I'll add them for `getSides`

src/padding.css
obartra: sure

PR #107, opened 2017-06-22
refactor(api): Improve consistency

obartra:
- Simplify CSS
- Treat margin like padding and allow for strings like "horizontal left" or "all"
- changed `margin` to `itemMargin` and `marginSize` to `itemMarginSize` on `Grid`
- `margin` and `marginSize` on Grid now refers to the margin of the `Grid` itself.
- `Layout`, `Grid`, and `Item` now support `margin` and `marginSize` with consisten parameters
- `Grid` additionally supports `padding` and `paddingSize`


src/types.js
```diff
@@ -18,9 +18,7 @@ export type AlignGrid = AlignItem | 'stretch'
 
 export type Justify = 'left' | 'center' | 'right' | void
 
-export type Margin = 'none' | 'all' | 'horizontal' | 'vertical' | void
```
arahansen: why are we getting rid of the Margin type? Seemed useful.

src/utils.js
```diff
@@ -73,6 +39,29 @@ export function getSides(sides?: string = ''): Array<IndividualSides> {
   return uniq(allSides)
 }
 
+export function getMarginClasses(
+  margin?: string,
+  marginSize?: MarginSizes
+): Array<string | void> {
+  const marginSizeClasses = {
```
arahansen: I think we should move this out so we aren't creating a new object essentially on every render

src/types.js
obartra: Now everything uses `IndividualSides` instead. "horizontal", now is the same than `"left right"`, `"left left right right"`, `"horizontal"`, `"horizontal right"` and we cannot account for all permutations. Because of this `Margin` type is no longer used.

I think this highlights the increased consistency of the API since now all margins and padding are treated the same way

src/utils.js
obartra: sounds like a problem for V8 to optimize, I'm happy to move it but I doubt that it makes any difference on runtime

PR #111, opened 2017-06-24
Build publish

### LINKED ISSUE #36: Refactor Build Process

obartra:
- Create `build` task:
  - The generated bundle requires React and PropTypes
  - The bundle is placed at `/dist/reflex.js`
  - Storybook uses the built bundle instead of `src/index`
  - Export only lodash modules used

- `yarn start` initiates 2 watch tasks (1 for `/src` and one for `/stories`)
- Transform flow to PropTypes for test, dev and prod
- Set up semantic release on CI for passing master builds

Closes #36

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,4 +1,53 @@
 {
-  "presets": ["es2015", "react"],
-  "plugins": ["transform-object-rest-spread", "transform-class-properties"]
+  "presets": [
+    "es2015",
+    "react"
+  ],
+  "plugins": [
+    "flow-react-proptypes",
+    "lodash",
+    "transform-object-rest-spread",
+    "transform-class-properties"
+  ],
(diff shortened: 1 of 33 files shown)


obartra:
Merging to test deployment and semantic release but there's still some work to be done

PR #120, opened 2017-06-26
refactor(margin): Standardize margins and paddings

obartra:
- standardize behavior between Grid and Item and margins and paddings.
- Grid / Item have different margin defaults but they are treated the same way.
- You can pass an array that defines the size, e.g. `[1, 1, 1, 1]` or `[1]` (they are the same).
  That would set the margin to 1 gutter on each direction. valid values are `0`, `0.5`, `1` and `2`

BREAKING CHANGE: margins and paddings switched to arrays


src/utils.spec.js
```diff
@@ -1,49 +1,65 @@
...
-    expect(all.includes('bottom')).toBeTruthy()
+describe('getSpacingClasses', () => {
+  it('should no expand 0 values', () => {
+    expect(getSpacingClasses([0])).toEqual([])
   })
 
-  it('should convert "horizontal" to left/right directions', () => {
-    const all = getSides('horizontal')
+  it('should expand single values', () => {
+    expect(getSpacingClasses([0.5])).toEqual([
```
arahansen: snapshot test could be useful here?

stories/core/marginTypes.js
```diff
@@ -0,0 +1,15 @@
...
+export const item = [0, 0.5, 1]
+export const none = []
+export const right = [0, 1, 0, 0]
+export const top = [1, 0, 0, 0]
+export const topLeft = [1, 0, 0, 1]
+export const topRight = [1, 1, 0, 0]
+export const vertical = [1, 0]
+export const verticalHalf = [0.5, 0]
+export const xBottom = [1, 1, 0]
+export const xItem = [1, 0.5, 0]
```
arahansen: what does xItem mean?

src/utils.js
```diff
@@ -1,12 +1,21 @@
...
-import marginStyle from './margin.css'
+import { memoize, compact } from 'lodash'
+import type { Spacing, Component } from './types'
+import style from './spacing.css'
 
 /* eslint-disable no-unused-vars */
 const noop = (...params: any[]) => null
 /* eslint-enable no-unused-vars */
 const isProd = process.env.NODE_ENV === 'production'
+const stringify = param => JSON.stringify(param)
```
arahansen: `const { stringify } = JSON`?

arahansen:
awesome tests 👍

src/utils.js
obartra: nice, I thought I would have to bind to JSON but I didn't try it, this looks cleaner, will update

stories/core/marginTypes.js
obartra: I'm using `xItem` and `xBottom` to mean "opposite of"

src/utils.spec.js
obartra: good call

stories/core/marginTypes.js
arahansen: hmm, its not super descriptive. But I'm not sure I have a better alternative. Maybe "AntiBottom"? "ReverseBottom"?

obartra: I'm fine with either, tbh I don't like any of them xBottom, antiBottom, reverseBottom all sound bad to me

arahansen: maybe "inverse"

arahansen: if none of those sound any better, might as well stick with x

PR #125, opened 2017-06-27
Kill item

### LINKED ISSUE #115: Simplify .box
Ideally `Box` would only set the text and the styling but everything else would be defined using `Item` properties.

That would better help demonstrate the behavior of `reflex` and reduce the amount of custom css

obartra:
- Remove `Item`
- Add `"auto"` as a `Grid` size
- Update tests to reflect changes
- Small test visual update

BREAKING CHANGE, Closes #115


obartra:
Right now `<Item/>` is just `<Grid margin={item} />`. I would say let's add a couple more mocks and see which patterns we can extract. I'm sure we'll find more than just Item as candidates for refactor

PR #135, opened 2017-06-28
refactor(Grid): Remove  Grid parameter

obartra:
with a few changes to the Layout Component, `<Grid root/>` can be implemented by setting
already existing properties.

To illustrate this usage case, index exports `Root` which is defined as:

```javascript
<Grid {...props} margin={[0, 0.5]} padding={[0, 1]} />
````

A similar element could be create for `Item` and `Row`

To avoid confusion, this renames the `Root` component in `/stories` to `RootLayout` since
that now wraps the new `<Root/>` component in a `<Layout type="parent"/>` Component

BREAKING CHANGE, closes https://github.com/obartra/reflex/issues/130

### CODE REVIEW

File: src/derived/root.js [hunk 1]
```diff
@@ -0,0 +1,7 @@
+// @flow
+import React from 'react'
+import Grid from '../grid'
+
+export default function Root(props: { [string]: any }) {
```
arahansen: props here should be GridProps maybe?

obartra: Only a subset since you shouldn't set margin or padding. It can def be better than this definition though 😆

PR #138, opened 2017-06-28
feat(testing): Add screenshot testing

obartra:
- Add screenshot testing
- Remove screener.io
- To update an image, delete from repo and CI will commit the additional images


obartra:
Yeah for sure it should probably have a dedicated wiki page as part of the development guide.

Right now the process is you remove any screenshots you don't want (or don't have for new stuff), then push to CI and it will create a commit for you on that same branch including the updated screenshots. Then we can verify them through the pull request

PR #142, opened 2017-06-30
docs(storybook): add search layout example

arahansen:
adds a more comprehensive example of a search results page; closer to "real world" usage


src/derived/col.js
```diff
@@ -0,0 +1,25 @@
...
+  marginBottom = 1,
+  marginLeft = 0.5,
+  ...props,
+  children,
+}: ColProps) =>
+  <Grid {...props} margin={[marginTop, marginRight, marginBottom, marginLeft]}>
+    {children}
+  </Grid>
+
+export default Col
```
obartra: 👍

src/index.js
```diff
@@ -8,4 +8,5 @@ export { default as LayoutHOC } from './layout.hoc'
 export { version } from '../version'
 export const utils = srcUtils
 
+export { default as Col } from './derived/col'
 export { default as Root } from './derived/root'
```
obartra: This is fine for now but I would like to have a better way to distinguish between core layout components and derived ones

storybook/stories/components/header/app.js
```diff
@@ -0,0 +1,95 @@
...
+            className={styles.colors2}
+          >
+            consectetur
+          </Col>
+        </Col>
+        <Col
+          style={{
+            borderLeft: '1px solid #fff',
+            borderRight: '1px solid #fff',
+          }}
```
obartra: why do you need this?

storybook/stories/components/header/app.js
```diff
@@ -0,0 +1,95 @@
...
+          marginBottom={0}
+          offset={1}
+          size={3}
+          className={styles.colors4}
+        >
+          <input
+            style={{ outline: 'none', padding: 0 }}
```
obartra: can you use the styling from `page` for input?

storybook/stories/components/header/app.js
```diff
@@ -0,0 +1,95 @@
...
+                backgroundColor: '#fff',
+                height: 15,
+                width: 15,
+                minHeight: 15,
+              }}
+            />
+          </Col>
+          <Col size="auto" marginBottom={0}>
+            <div
+              style={{
```
obartra: I think I would rather move the custom css to use class names so it doesn't distract from the layout

storybook/stories/components/search/filters.js
```diff
@@ -0,0 +1,28 @@
+// @flow
+import React from 'react'
+import { Col, Grid } from 'reflex'
+import styles from '../../../shared/stories.css'
+
+export default function SearchFilters() {
+  return (
+    <Col size={3} className={styles.colors4} align="top">
```
obartra: This size is specific to the search page, I think it would make more sense to define a single Filter than a sized column of filters. That way you can put the column with the rest of the layout but reduce the boilerplate

storybook/stories/components/search/nav.js
```diff
@@ -0,0 +1,62 @@
+// @flow
+import React from 'react'
+import { Col, Layout, Grid, Root } from 'reflex'
+import styles from '../../../shared/stories.css'
+
+export default function SearchNav() {
+  return (
+    <Layout className={styles.colors2} margin={[1, 0, 0, 0]}>
```
obartra: Actually just seeing this made me wonder, do we need margin / padding for layout? Can we implement the current designs without it at all?

storybook/stories/components/search/results.js
```diff
@@ -0,0 +1,100 @@
...
+
+export default function SearchResults({
+  results = 1,
+  pages = 4,
+}: {
+  results: number,
+  pages: number,
+}) {
+  return (
+    <Grid size={9} className={styles.colors2} align="top">
```
obartra: Same thing than with filters, this is specific to the search page, I think it would make sense to put here a single search result and have the search page use it.

obartra:
Awesome job with the new `Col` element, I think it will be really useful.

My only concern is with the division of components. I agree that for a real world scenario we may want to have the divisions you have (and then individual components as well) but for the stories in storybook I would like each component to be independent

storybook/stories/components/search/filters.js
arahansen: I agree, and had considered it. I didn't go for it since I thought it would make the independent story look awkward.

But having a flexible, independent story seems more compelling and useful

src/index.js
arahansen: I don't have a problem keeping them top level, since they're going to be very common primitives.

storybook/stories/components/header/app.js
arahansen: it's specific to the comp I was working off of. I can remove it.

storybook/stories/components/header/app.js
arahansen: for demo purposes, it'll work

storybook/stories/components/search/nav.js
arahansen: meaning we could just be using a nested Grid?

obartra: hm could this be implemented with Grid padding instead of nesting the Grid? If the answer is no then I agree there is still a need for Layout margin

storybook/shared/stories.css
```diff
@@ -51,26 +54,51 @@ button {
   background-color: var(--axonGold);
 }
 
-.box1, .box2, .box3, .box4 {
```
obartra: is this caused by different prettier versions?

storybook/stories/examples/search.js
```diff
@@ -0,0 +1,42 @@
...
+export default function() {
+  const results = number('Results', 5, { range: true, min: 1, max: 10 })
+  const pages = number('Pages', 3, { range: true, min: 1, max: 5 })
+  return (
+    <Layout type="parent" className={styles.page}>
+      <Header />
+      <Layout
+        type="stretch"
+        className={styles.content}
+        style={{ marginTop: 40 }}
```
obartra: whaaaat?

storybook/shared/stories.css
arahansen: is it possible prettier hasn't been run on this file before?

storybook/stories/examples/search.js
obartra: this should be `marginTop={2}` if anything but probably just `marginTop{1}` def no custom margins

storybook/shared/stories.css
obartra: no

PR #145, opened 2017-06-30
test(image): Co-locate images

obartra:
Use `storyFolders` to get the image names and colocate test images. This should make updating them easier


storybook/shared/storyFolders.js
```diff
@@ -28,6 +28,24 @@ function getFilesAndFolders(
...
 }
 
+function getImage(files, filepath) {
+  const pngFile = filepath.replace(/\.js$/, '.spec.png')
+  const hasPNG = files.indexOf(pngFile) !== -1
+
+  return hasPNG ? pngFile : ''
+}
+
+function dropEnds(array) {
```
arahansen: `const dropEnds = flow(tail, initial)`

obartra: yeah I rather avoid importing `flow` for this one case

arahansen: Are we concerned with bundle size for a dev tool?

obartra: oops good point we are not, I totally missed that, will update

PR #151, opened 2017-07-02
feat(offset): remove offset as a prop for Grid

arahansen:
offset is a nice-to-have but ultimately not necessary to implement all layouts we have come across
thus far. Removing offset will open the door for allowing the margin prop to use css margins

remove offset prop functionality

NOTE: it does seem to make some layouts slightly more awkward to produce. Usually requiring some sort of nested `Grid`/`Col` combo.


src/types.js
```diff
@@ -2,8 +2,6 @@
 
 type zeroThroughEleven = 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11
```
obartra: this is just used by `Size` now, we may as well move it there

storybook/stories/api/offset.js
```diff
@@ -1,39 +0,0 @@
-// @flow
```
obartra: Do you think there is value in keeping this example but updating to use `<Offset />`?

storybook/stories/components/card/filter.js
```diff
@@ -69,7 +74,9 @@ export default function() {
                 <ListItem index={4} />
 
                 <Grid>
-                  <p>{loremIpsum.substr(0, 100)}</p>
+                  <p>
+                    {loremIpsum.substr(0, 100)}
+                  </p>
```
obartra: have we verified if these chages are due to prettier version differences?

storybook/stories/components/header/app.js
```diff
@@ -7,7 +7,7 @@ export default function() {
   return (
     <Layout fixed="top" className={styles.colors1}>
       <Root align="center">
-        <Col align="center" size="auto" marginBottom={0}>
```
obartra: why is `"auto"` wrong here?

storybook/stories/components/header/app.js
```diff
@@ -49,8 +49,15 @@ export default function() {
...
         </Col>
-        <Col marginBottom={0} offset={1} size={3} className={styles.colors4}>
-          <input placeholder="Search for anything..." />
+        <Col
+          marginBottom={0}
+          justify="right"
+          size={3}
+          className={styles.colors4}
+        >
+          <Grid>
```
obartra: No need for `Grid` here, is there?

storybook/stories/components/card/filter.js
arahansen: Yeah I'm pretty sure my home laptop has a different version of the prettier plugin. Although I thought we had a pre commit hook that ran prettier? Which I would expect to be the same version for the project?

obartra: hm yeah that's weird, let's not worry about it for now we can tackle it later

storybook/stories/api/offset.js
obartra: Yeah I think there's value to have it, particularly given that most 12 column grid systems offer it

PR #153, opened 2017-07-03
feat(dev-mode): adds devMode flag to Layout for enabling colored Grid backgrounds

arahansen:
Provides useful debugging tool that clearly shows the boundary of Cols and Grids. Usage is
demonstrated with Storybook Dev Mode toggle.

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,7 +90,7 @@
     ]
   },
   "scripts": {
-    "build": "webpack --config scripts/webpack.config.js -p",
+    "build": "yarn webpack -- -p",
     "commit": "git-cz",
     "contributors:add": "all-contributors add",
     "contributors:gen": "all-contributors generate",
@@ -103,10 +103,11 @@
     "precommit": "lint-staged",
     "semantic-release": "semantic-release pre && npm publish && semantic-release post",
     "start:storybook": "start-storybook -p 9001 -s ./storybook/static -c storybook/config",
-    "start:webpack": "yarn build -- --watch",
+    "start:webpack": "yarn webpack -- --watch",
     "start": "npm-run-all --parallel start:*",
     "test:image": "babel-node node_modules/.bin/nightwatch --config test/screenshot/nightwatch.conf.js",
-    "test": "jest"
+    "test": "jest",
+    "webpack": "webpack --config scripts/webpack.config.js"
   },
   "release": {
     "verifyConditions": "condition-circle",
--- a/scripts/webpack.config.js
+++ b/scripts/webpack.config.js
@@ -1,6 +1,7 @@
 const CleanWebpackPlugin = require('clean-webpack-plugin')
 const { BundleAnalyzerPlugin } = require('webpack-bundle-analyzer')
 const { resolve } = require('path')
+const { DefinePlugin } = require('webpack')
 
 const isProd = process.env.NODE_ENV === 'production'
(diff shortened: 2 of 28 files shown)


package.json
```diff
@@ -103,10 +103,11 @@
     "precommit": "lint-staged",
     "semantic-release": "semantic-release pre && npm publish && semantic-release post",
     "start:storybook": "start-storybook -p 9001 -s ./storybook/static -c storybook/config",
-    "start:webpack": "yarn build -- --watch",
+    "start:webpack": "yarn webpack -- --watch",
```
obartra: with this change it won't run webpack with production settings. Because we are using different babel aliases and the build time is short no matter what, I think it's worth keeping it running with `-p` even on dev mode

obartra: Oooh but the whole point is that we are testing dev mode now lol, hmmm maybe we can set the babel aliases for dev and prod to match, then 👍

scripts/webpack.config.js
```diff
@@ -28,6 +29,11 @@ module.exports = {
       logLevel: 'error',
       openAnalyzer: false,
     }),
+    new DefinePlugin({
+      'process.env': {
+        NODE_ENV: JSON.stringify(isProd ? 'production' : 'development'),
```
obartra: this should probably be `JSON.stringify(process.env.NODE_ENV || 'development')` instead

src/grid.hoc.js
```diff
@@ -17,29 +20,39 @@ export type Props = {
...
+    }
+
     render() {
       const {
+        dev,
         align,
         children,
         className,
         justify,
-        margin = this.context.margin,
```
obartra: Good catch, this was always undefined right?

storybook/shared/withExtensions.js
```diff
@@ -1,15 +1,21 @@
 // @flow
 import React from 'react'
+import PropTypes from 'prop-types'
 import { WithNotes } from '@storybook/addon-notes'
 import { boolean } from '@storybook/addon-knobs'
 import style from './designGrid.css'
 
-let showOverlay = false
+// let showOverlay = false
```
obartra: oops

storybook/shared/withExtensions.js
```diff
@@ -24,9 +30,8 @@ export default class WithExtensions extends React.PureComponent {
 
   render() {
     const { notes, className, ...props } = this.props
-
     // defaults to last used value, when changed it updates it. This allows us to persist the 'show overlay' setting across examples
-    showOverlay = boolean('Overlay', showOverlay)
+    const showOverlay = boolean('Overlay', false)
```
obartra: Hm is 'overlay' still a good name? we could call it 'dev mode'

obartra:
Changes look great, it may be worth updating `Box.js` to also use dev mode internally (based on the `type`)

src/grid.hoc.js
arahansen: I believe so, haha

storybook/shared/withExtensions.js
arahansen: I think overlay still works

arahansen:
@obartra codebeat's wrong. This should be merged 😂

obartra:
lol I was actually just looking for this thinking it was already merged. Merging and giving you admin rights

PR #154, opened 2017-07-03
fix(publish): include dist files for npm publish

arahansen:
Was missing a `files` entry in package.json which denotes which files should be published to npm

new files available on npm

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
   "main": "dist/reflex.js",
   "browser": "dist/reflex.js",
   "homepage": "https://obartra.github.io/reflex",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -6,6 +6,7 @@
   "main": "dist/reflex.js",
   "browser": "dist/reflex.js",
   "homepage": "https://obartra.github.io/reflex",
+  "files": ["package.json", "README.md", "LICENSE", "src", "dist"],
```
obartra: do we want to include `src` in the published package? only `"dist"` should be needed unless you want it as reference

arahansen: I like including it for reference

PR #156, opened 2017-07-05
test(warnings): Reduce warnings when running tests

obartra:
`yarn test` was showing some unnecessary warnings. Fixes related issues (by adding keys)


arahansen:
😳 👍

PR #157, opened 2017-07-06
fix(api): removed hoc wrappers

arahansen:
HOCs made more sense when we wanted this library to act as wrappers for other components. The
philosophy of the library has changed since then, and HOCs don't make sense any longer. They were
causing bugs through the use of PureComponent + context


src/designGrid.css
```diff
@@ -0,0 +1,63 @@
+@import "./variables.css";
```
obartra: Should this replace the existing design Grid on the story examples?

src/grid.hoc.js
```diff
@@ -1,85 +0,0 @@
-// @flow
```
obartra: I'm in favor of removing `layout.hoc`. For `grid.hoc` I can see it both ways so this is fine. My rationale is that GridHOC may be used in styleguide components. Ultimately though it may be easier to think of this library as just 2 components and then wrap anything on your styleguide with one of them

obartra:
My only concern is on duplicating the design Grid. I think we should update all examples to only use this one (replacing overlay with just devMode). everything else 💯

src/designGrid.css
arahansen: uh oh, I think this accidentally got pushed with an earlier change I was playing with?

obartra: oh got it, it's a pity, it would have been nice to have 😆 

Then yeah remove it and  then it's ready to merge, approving the PR now

PR #158, opened 2017-07-08
ci(build): Measure build size

obartra:
Add build size check to prevent the gzipped file size of the bundle to grow over 9kb


obartra:
bithound is complaining about eslint which we can't upgrade until https://github.com/airbnb/javascript/issues/1447 is resolved

obartra:
merging since it's just ci changes

PR #159, opened 2017-07-08
docs(readme): Remove bettercodehub check

obartra:
Remove Better Code Hub check because it has several issues with JS that don't make it relevant. E.g.
it considers classes with more than 15 lines problematic but a React class with just an empty render
and a few props already exceeds it


obartra:
merging, documentation only changes

Issue #100, opened 2017-06-20
grid on small resolution breaks


obartra:
Grid overlay doesn't work correctly when the window is small.

Some amount of distorsion is expected because once elements hit their minimum size they will no longer align to the grid. That's not the issue.

The issue is with how the grid becomes a gradient because it's incorrectly defined. It looks like this:

![image](https://user-images.githubusercontent.com/3877773/27342290-a2ef9cc6-5594-11e7-8190-46eb2188e56a.png)

PR #160, opened 2017-07-09
fix(dev): Improve overlay grid

obartra:
- Fix a bug where column width wasn't computed correctly
- Chrome re-renders the gradient background less often than the actual content
  which results in inconsistencies when the page is smaller than the max content
  size (1224px), after that the content doesn't vary (it's just centered) so
  Chrome renders the right background grid.
- To show the design grid but discourage usage, when the window width is less than
  1224px, the grid is shown in yellow (which makes it harder to view the edges)
- The Chrome issue can be seen by slowly resizing the window width. At specific pixel sizes it works as expected but then falls behind until the next change (~10px intervals)

Closes https://github.com/obartra/reflex/issues/100


obartra:
This only fixes the design grid but doesn't make it part of the dev tools yet

Issue #143, opened 2017-06-30
Prepend Github source to md file documentation


arahansen:
All notes for storybook examples should include a link to the perspective full sources on github.

Example:

# Example 1
This example shows how to do the dougie.

# Link to Full Source:
[ link to github dougie.js ]

PR #161, opened 2017-07-09
docs(notes): Include link to file

obartra:
Includes link to source file on each example

Closes https://github.com/obartra/reflex/issues/143

image


package.json
```diff
@@ -20,7 +20,7 @@
     "@bubltechnology/customizable-commit-analyzer": "1.0.2-0",
     "@storybook/addon-knobs": "3.1.6",
     "@storybook/addon-notes": "3.1.6",
-    "@storybook/addon-storyshots": "3.1.7",
+    "@storybook/addon-storyshots": "3.1.6",
```
arahansen: was there a bug in 3.1.7?

obartra: no, this is just so they are all on the same version but we still have that same issue with the latest one

PR #169, opened 2017-07-11
fix(stories): update invalid stories

arahansen:
The babel plugin that converts flow types to prop-types was correctly throwing an error.
However, due to a bug in that babel plugin, information about the invalid prop-types were
being obscured.

There is a PR open for this babel plugin that should correct this issue: https://github.com/brigand/babel-plugin-flow-react-proptypes/pull/116

### DIFF
--- a/src/types.js
+++ b/src/types.js
@@ -3,7 +3,7 @@ type zeroThroughEleven = 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11
 
 export type Offset = zeroThroughEleven | void
 
-export type Size = zeroThroughEleven | 'fit' | 'auto' | void
+export type Size = zeroThroughEleven | 12 | 'fit' | 'auto' | void
 
 export type Component =
   | {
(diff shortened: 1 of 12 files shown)


src/types.js
```diff
@@ -3,7 +3,7 @@ type zeroThroughEleven = 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11
 
 export type Offset = zeroThroughEleven | void
 
-export type Size = zeroThroughEleven | 'fit' | 'auto' | void
+export type Size = zeroThroughEleven | 12 | 'fit' | 'auto' | void
```
arahansen: I know we had discussed removing `12` as an option, but it's technically a valid prop value and is causing prop-types to invalidate when storybook knobs is passing `12`. The other option would be to have the storybook knobs range from 1 to 11, or undefined (if thats possible).

obartra: Do we need `zeroToEleven` for anything? if not, let's consolidate all into a single type (`zeroToTwelve`).

obartra: disregard my comment above. Also I agree woth you size should allow 12

obartra: disregard my comment above. Also I agree woth you size should allow 12

obartra: disregard my comment above. Also I agree woth you size should allow 12

PR #172, opened 2017-07-12
build(css): extract css into standalone bundle

arahansen:
BREAKING CHANGE requiredcss moved into separate file


arahansen:
Yeah, I was gonna update this pr with the new requirements.

Probably update the minimum example, and a note about how we are doing CSS.

obartra:
Sounds good, after you add these changes feel free to merge 👍

PR #173, opened 2017-07-12
docs(examples): add link to code sandbox editable demo

arahansen:


obartra:
looks like we got another false positive on CI but changes look good, ready to merge!

Issue #174, opened 2017-07-13
Fix devMode examples


obartra:
Most examples don't respond to devMode flag because they have set backgrounds, update!

PR #176, opened 2017-07-15
refactor(examples): Better devMode usage

obartra:
- Refactor so usage examples need less custom css
- Ensure devMode is used by all examples

Closes https://github.com/obartra/reflex/issues/174


storybook/shared/box.js
```diff
@@ -8,39 +8,34 @@ const typeMap = {
...
-          {children || value}
-        </Grid>
+export default function Box({
+  children,
+  margin = [0, 0.5, 1],
+  style,
+  type,
+  value = type,
+  ...props
+}: {
```
arahansen: This might be good to pull outside the function parameters. A bit hard to parse when its inlined like this.

storybook/stories/examples/cards.js
```diff
@@ -3,7 +3,7 @@ import React from 'react'
 import { Grid, Layout, Root } from 'reflex'
 import { Box } from '../../shared'
 import styles from './layout.css'
-import { top, horizontalHalf } from '../../shared/marginTypes'
```
arahansen: hmm why did the screenshot pic so subtly change here? is that significant enough to look into?

obartra: it's intentional. Things were relying on hardcoded margins and I replaced all these by actual reflex margin properties. So they don't match exactly but I think this is a better use of them.

E.g. in one instance `h1` had a `10px 0` margin so I replaced that with `verticalHalf` which is `12px` instead of `10px`

Issue #175, opened 2017-07-13
Webpack Includes `eval` check in dist


arahansen:
When bundling, webpack is including an `eval` check to see if its available.

This causes a warning for servers that have no eval CSP settings enabled.

We should explore if it's possible to remove this eval from the bundle to remove this console warning.

arahansen:
Code in question:

```
try {
	// This works if eval is allowed (see CSP)
	g = g || Function("return this")() || (1, eval)("this");
} catch (e) {
	// This works if the window reference is available
	if ((typeof window === "undefined" ? "undefined" : _typeof(window)) === "object") g = window;
}
```

obartra:
Yeah it's annoying. It causes it to report the issue but it wouldn't break anything since it's just testing if CSP is enabled or not. I don't know if I would count it as an issue though

obartra:
Also, could this be related to the sourcemaps? I'm wondering if this happens with the prod build

obartra:
This is the offending file https://github.com/webpack/webpack/blob/master/buildin/global.js

obartra:
Looks like redux is not using webpack anymore but when upgraded to webpack 2 this is how they tackled it: https://github.com/reactjs/react-redux/pull/680/files#diff-11e9f7f953edc64ba14b0cc350ae7b9dR20

PR #178, opened 2017-07-16
build(csp): Remove eval from bundle

obartra:
- Webpack 2 includes a CSP check which causes an (intended) console error.
- To prevent that (and assume CSP is always enabled) these changes follow
  the same pattern than redux and replace it from the build
- Remove comments from the css minified bundle
- Use a single helper definition for babel instead of including it per file

Closes https://github.com/obartra/reflex/issues/175

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
     "npm-run-all": "4.0.2",
     "postcss-cli": "4.1.0",
     "postcss-cssnext": "3.0.0",
+    "postcss-discard-comments": "2.0.4",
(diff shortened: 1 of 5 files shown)


arahansen:
ah, clever! 

I'm not sure I quite follow what `apply` is doing. Maybe a different naming convention for the function? or a brief description of why its included?

obartra:
Yeah I was thinking we can move it to a separate file and we treat it exactly like any other plugin

PR #183, opened 2017-07-18
docs(layout): Add additional layout examples

obartra:
Add layout examples showcasing the usage of different parameters akin to
how we do for Grid.


obartra:
Just documentation changes, merging

PR #184, opened 2017-07-19
feat(spacing): add spacing options for Grid and Layout

arahansen:
- adds margin{Top, Right, Bottom, Left} and padding{Top, Right, Bottom, Left} properties to `Grid` and `Layout` (margin only on Layout).
- If margin props and margin value props are included, an error will be thrown in development.
- This PR should also update some examples to use this new syntax.


obartra:
Did the bundle size increase?

src/grid.js
```diff
@@ -1,21 +1,36 @@
...
   padding?: Spacing,
   size?: Size,
+  marginTop?: SpacingValues,
+  marginRight?: SpacingValues,
+  marginBottom?: SpacingValues,
+  marginLeft?: SpacingValues,
+  paddingTop?: SpacingValues,
+  paddingRight?: SpacingValues,
+  paddingBottom?: SpacingValues,
+  paddingLeft?: SpacingValues,
```
obartra: this is reeeally nit picky but you could sort them alphabetically with something like https://github.com/Tyriar/vscode-sort-lines

src/layout.js
```diff
@@ -61,24 +61,26 @@ export default class Layout extends React.Component {
     margin?: Spacing,
     overflow?: Overflow,
     type?: LayoutType,
+    children: React$Element<*>,
```
obartra: Why is children needed? It looks like it's treated like any other property

src/utils.js
```diff
@@ -30,6 +33,7 @@ export function getDisplayName(WrappedComponent: string | Component): string {
 function getSizeName(size: number) {
   switch (size) {
     case 0:
+    case undefined:
```
obartra: Undefined is "single" in terms of class names, should not be 0 unless you update the css

src/utils.js
```diff
@@ -77,4 +112,7 @@ function getSpacing(
   )
 }
 
-export const getSpacingClasses = memoize(getSpacing, stringify)
+export const getSpacingClasses = memoize(
+  getSpacing,
+  (props: GridProps, type: string) => type + stringify(props)
```
obartra: This is fine but I expected `eslint` would yell at you for not interpolating

arahansen:
closing in favor of https://github.com/obartra/reflex/pull/189

PR #189, opened 2017-07-21
feat(spacing): add top/right/bottom/left padding/margin options

arahansen:
- Layout supports margin{Top, Right, Bottom, Left}
- Grid now supports the same for margin and padding
- updated examples to make use of new functionality


arahansen:
I recall that we wanted to explicitly error on this case. I could see it as confusing keeping track which property gets priority.

obartra:
I agree that the case `margin={[0, 0.5, 1, 2]} marginTop={2} marginRight={1}` would be confusing. I still think there may be value at combining them and leave it up to the user if they choose to make it ambiguous. The pattern would be very clear (`marginDirection` always overrides) and similar to doing something like:

```css
.class {
  padding: 0 1px;
  padding-left: 3px;
}
```

I could go either way though.

If I'm reading this implementation correctly, it silently ignores `marginDirection` and `paddingDirection` if `margin` or `padding` (respectively) are set. I don't think we want that, we should throw an error if we want to prevent this behavior.

src/utils.js
```diff
@@ -1,6 +1,8 @@
 // @flow
-import { memoize, compact } from 'lodash'
+import { memoize, compact, pick, capitalize, every, find } from 'lodash'
```
obartra: Do we need all these additional methods? They are increasing the bundle size

src/grid.js
```diff
@@ -15,18 +22,28 @@ export type Props = {
...
-    margin: [],
-  }
-
   static contextTypes = {
     devMode: PropTypes.bool,
   }
 
+  componentWillReceiveProps(newProps: Props) {
+    validateSpacingProps(newProps, 'margin')
+    validateSpacingProps(newProps, 'padding')
```
obartra: Oh I see now you are applying the validation separately. I think we should do this as part of setting the spacing classes. I don't think it's the responsibility of this component to set everything and then check it was the right thing

src/utils.js
arahansen: Yeah definitely not. I was hoping they wouldn't increase the bundle as much as they did :(

src/grid.js
arahansen: Yeah that'll work.

src/utils.js
```diff
@@ -32,18 +39,18 @@ export function getDisplayName(WrappedComponent: string | Component): string {
...
   type: 'margin' | 'padding'
 ) {
   if (process.env.NODE_ENV !== 'production') {
-    const values = pick(props, directions.map(dir => type + capitalize(dir)))
-
-    // flow complains about props since `padding` cannot be found on `LayoutProps`
-    const value = (props: any)[type]
+    const values = getSpacingValues(props, type)
+    // flow is having trouble understanding what `props[type]` could be
+    const value: any = props[type]
```
obartra: can we cast it? something like `const (value: any): SpacingValues = props[type]`. That way we don't need the comment and flow knows what it's getting

Issue #144, opened 2017-06-30
Generate flow types on production build


arahansen:
When building a deployable bundle, a `.flow` file should be generated and output in the same directory as the main bundle. 

This will allow users that have flow enabled for their projects to benefit from the types we have already defined.

PR #192, opened 2017-07-23
feat(flow): add flow types to dist

arahansen:
- adds build step to copy src files and output to dist as `*.flow`. These files will be picked up by `flow` when type checking
- fixed an issue with flow config where flow was looking at dist for `reflex` components. Although, this could theoretically be resolved by the new build step implemented here.

Resolves https://github.com/obartra/reflex/issues/144

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -32,7 +32,7 @@
           "module-resolver",
           {
             "alias": {
-              "reflex": "./src/index"
+              "reflex": "./src/reflex"
             }
           }
         ]
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,4 +1,5 @@
 [ignore]
+<PROJECT_ROOT>/dist
 .*/node_modules/stylelint/.*
 .*/node_modules/eslint-plugin-jsx-a11y/.*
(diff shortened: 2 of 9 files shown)


arahansen:
Unfortunately not from what I can tell. :(
https://flow.org/en/docs/types/modules/#importing-and-exporting-types-a-classtoc-idtoc-importing-and-exporting-types-hreftoc-importing-and-exporting-typesa

storybook/stories/components/search/nav.js
```diff
@@ -4,7 +4,7 @@ import { Col, Layout, Grid, Root } from 'reflex'
 
 export default function SearchNav() {
   return (
-    <Layout dev={4} margin={[1, 0, 0, 0]}>
+    <Layout dev={4} margin={[3, 1, 4, 0]}>
```
arahansen: yep, should be reverted now

obartra:
Yeah I was just looking at that. oh well don't worry about it. I do see some linting issues on flow types, maybe we should exclude that folder from linting on bithound/cc/eslint?

arahansen:
Linting issues should be resolved.

PR #199, opened 2017-07-25
build(flow): strip content from *.css.flow files

arahansen:
- flow was trying to parse the contents of the *.css.flow files. Since we don't care about type info from these files anyways, we might as well just use stubs for their contents so flow doesn't complain about not being able to find the modules.


arahansen:
seems this will be tough to verify on CI, but I was able to verify locally on a new CRA project that type definitions are working properly and the css stubs are treated as such using `yarn build && ./scripts/generateCssFlow.js`

scripts/generateCssFlow.js
```diff
@@ -0,0 +1,19 @@
...
+  throw new Error('dist folder not available')
+}
+
+const files = fs.readdirSync(distDir)
+
+files.forEach(file => {
+  const parsed = path.parse(file)
+  if (parsed.name.endsWith('.css')) {
+    const fullPath = path.resolve(__dirname, '../dist', file)
+    console.log('stubbing css file', fullPath)
```
obartra: I we are adding JS to `scripts` we should include it in the linting config. This should complain about console. You can add `eslint-disable no-console` or just use ``process.stdout.write(`stubbing css file ${fullPath}\n`)``

scripts/generateCssFlow.js
```diff
@@ -0,0 +1,19 @@
...
+    fs.writeFileSync(fullPath, '')
+  }
+})
```
obartra: This could be a bit simpler with

```javascript
function isCSSFile(file) {
  return path.parse(file).endsWith('.css') && !fs.lstatSync(file).isDirectory()
}

function getTargetPath(file) {
  return path.resolve(__dirname, '../dist', file)
}

function writeEmptyFile(path) {
  return fs.writeFileSync(path, '')
}

fs
  .readdirSync(distDir)
  .filter(isCSSFile)
  .map(getTargetPath)
  .forEach(writeEmptyFile)
```

PR #201, opened 2017-07-26
feat(col): Allow margin prop

obartra:
margin prop was not allowed in Col because we were setting margin direction
properties and both cannot be set at the same time.

- Not set margin defaults when `margin` prop is passed
- Replace thrown errors with `console.error` to prevent the page from breaking

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -28,6 +28,10 @@ export function getDisplayName(WrappedComponent: string | Component): string {
 }
 
 export function validateSpacingProps(props: SpacingProps) {
+  if (isProd) {
+    return true
+  }
+
   const margins = ['marginTop', 'marginRight', 'marginBottom', 'marginLeft']
   const paddings = [
     'paddingTop',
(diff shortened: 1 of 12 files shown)


src/utils.js
```diff
@@ -28,6 +28,10 @@ export function getDisplayName(WrappedComponent: string | Component): string {
 }
 
 export function validateSpacingProps(props: SpacingProps) {
+  if (isProd) {
```
arahansen: any idea if this is enough for uglify to dead code eliminate the rest of this function?

obartra: yeah this is removed out of `reflex.min.js` but not from `reflex.js` (since `reflex.js` is not uglified)

arahansen: awesome!

obartra: There are some additional optimizations we could do for reflex but when I tried it we were getting 0.05kb more so I don't think it's worth it

PR #207, opened 2017-07-28
feat(grid8): Update Base to take a base unit

obartra:
With base now it's easy to create additional grids with different bases.

Grid8 would allow us to implement things like small cards without either having
to directly set css or use imprecise fractions

### DIFF
--- a/src/derived/grid8.js
+++ b/src/derived/grid8.js
@@ -1,7 +1,0 @@
-// @flow
-import React from 'react'
-import Base from '../base'
-
-export default function Grid(props: {}) {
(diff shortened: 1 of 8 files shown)


src/derived/grid8.js
```diff
@@ -0,0 +1,7 @@
+// @flow
```
arahansen: is it really worth having an entire core API component for grid8? Its so so simple that it could just be left to the consumer to create a reusable component.

obartra: That's a good point, we could simply just export this updated `<Base>` and no need to add an additional component. I'll update

PR #212, opened 2017-08-02
test(sauce): Add IE11 SauceLabs image comparison tests

obartra:
🎉 Add support for IE 11 and fix bugs to add compatibility 🎉 

WIP: Behavior as expected, SauceLabs different browser windows strike again

| IE11 | Chrome | Diff |
|-----|---------|-----|
|  |  |  |

### DIFF
--- a/package.json
+++ b/package.json
@@ -141,7 +141,7 @@
     "start:webpack": "yarn build:dev -- --watch",
     "start": "npm-run-all --parallel start:*",
     "test:bithound": "bithound check git@github.com:obartra/reflex.git",
-    "test:image": "nightwatch --config test/screenshot/nightwatch.conf.js --env chrome,firefox,edge,safari",
+    "test:image": "nightwatch --config test/screenshot/nightwatch.conf.js --env chrome,firefox,edge,safari,ie11",
     "test:size": "bundlesize",
     "test": "jest",
     "test:watch": "jest --watch"
--- a/src/base.css
+++ b/src/base.css
@@ -44,7 +44,9 @@
   }
 
   &.col-fit {
-    flex-basis: 0%; /* % is required for Edge */
+    flex-grow: 0;
+    flex-shrink: 0;
+    width: auto;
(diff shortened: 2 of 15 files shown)


arahansen:
I noticed that we were targeting grandchildren for the max-width of non-layout elements. this was messing with the percentage-based max widths on ie11. 

Also, it seems `.wrapper` doesn't exist anymore, so pulled that out too.

If we have a good reason to be targeting grandchildren, feel free to back out my commit. But I didn't immediately see any breaking changes from removing that selector.

obartra:
ah yes I missed that, it's still a WIP, I need to update the images on the PR. The stretch problem is fixed. Current issues are incorrect width calculation and layout type parent height set at 100%.

The layout one doesn't have impact on the children and may be the only way to get cross browser consistent behavior with stretch.

The width issue is the blocking one since things overflow differently on IE11 with current changes

I think we're getting pretty close

arahansen:
I'm not convinced my change will work across the board. But it seemed to do the trick on the search page example.

Without the fix:
capture


With the fix:
capture

notice, for example, how "First View | Second View" is collapsed practically off the screen without the fix, but the width is set correctly with the fix in place

obartra:
That's fine, I'm still trying to figure out the SauceLabs issues, I would say let's split the work into 2 branches, that way we can both keep looking into it

obartra:
There are some cross browser issues with "Search" and "Two Sections" examples. They are related to the IE11 workaround for min-height.

SauceLabs window size issues are resolved, onward!

arahansen:
Looks like the issue is the footer showing up when the content should be pushing below the fold?

obartra:
yup. The trick is in solving this without hacks or extra dom markup in a way that works cross browser. Now it works on IE11 but the fix for Safari breaks it for chrome and vice versa. Open to suggestions

obartra:
Closing in favor of https://github.com/obartra/reflex/pull/227

PR #224, opened 2017-08-14
feat(api): add hoc wrappers

arahansen:


package.json
```diff
@@ -113,7 +113,7 @@
   "bundlesize": [
     {
       "path": "./dist/reflex.min.js",
-      "maxSize": "6.0 Kb"
```
obartra: Maybe 6.5?

src/layout.hoc.js
```diff
@@ -0,0 +1,132 @@
+// @flow
```
obartra: Is a layout hoc needed?

arahansen: I figured it'd be nice to have. But I don't really have a use case for it. So I'm fine dumping it

package.json
arahansen: I was going to, but readme looks like its already set to 7.0kb which is why I didn't change it. Also why I went with 7k for the maxSize

arahansen:
@obartra backed out layout changes.

package.json
obartra: it's +1 for the css size

arahansen: gotcha

PR #241, opened 2017-08-31
feat(overlay): Add dev only grid overlay

### LINKED ISSUE #230: Export Grid Overlay
Consider exporting grid overlay

It may be useful for packages consuming reflex to know if they are part of the Grid or not. It could behave similarly to `dev` mode and be stripped out automatically from prod builds.

If we do that we may also want to consider revamping it, maybe having a way to show the subgrid for each grid element

obartra:
To toggle overlay, press `ctrl+shift+k` or `cmd+shift+k`. This will toggle the overlay Grid. It has 3 components:

- 12 semi transparent Top level columns and page margins
- 1px gray outline on every `<Base>` component
- 12 black dots indicating each column within each `<Base>` Component subgrid

To toggle the color mode, press `ctrl+shift+s` or `cmd+shift+s`. This will toggle colors set with `dev` property on and off.

Note that unless you call `initDevMode` with `force` (`initDevMode({ force: true })`) it will only be enabled on development.

Current rendering looks like this:

image
image

Closes #230


src/dev.js
```diff
@@ -0,0 +1,47 @@
+// @flow
+import styles from './dev.css'
```
arahansen: why append to the body manually? any reason to not just do another `ReactDOM.render()` here?

obartra:
`CTRL+K` or `CMD+K` may collide with some apps, even if it only applies on dev mode.

Maybe we should change it to `CTRL+SHIFT+K` or `CMD+SHIFT+K`, what do you think?

obartra:
also todo this should work with custom bases

arahansen:
What about setting a default, and making it configurable via some sort of `configDevMode` function?

obartra:
I just added some more options, with this reflex would export an `initDevMode` function. You can pass additional parameters to customize the keybindings but it would otherwise default to adding listeners to `ctrl+shift+k` and `ctrl+shift+s` to toggle overlays or backgrounds.

These changes would only apply when not on production and they remove the need for setting `devMode` through context. Ultimately, the goal is to make it easier for people to actually use dev mode.

Issue #242, opened 2017-08-31
Add support for no arrays


obartra:
It would be convenient to be more flexible with input formats, particularly for margin and padding but also other props we may add (e.g. `show`)

Currently, this is the only valid syntax:

```jsx
<Grid margin={[0]} />
```

We should also support:

```jsx
<Grid margin={0} />
```

But we could also support

```jsx
<Grid margin="0" />
```

And then

```jsx
<Grid margin="0, 1, 2" />
```

or

```jsx
<Grid margin="0 1 2" />
```

Both of which may look a bit cleaner than the current:

```jsx
<Grid margin={[0, 1, 2]} />
```

PR #243, opened 2017-09-01
feat(base): Allow mixed spacing params

obartra:
- Add support for more input formats than arrays for spacing properties (margin / padding)
- Allows the following inputs https://regex101.com/r/bs73rZ/1/

Closes https://github.com/obartra/reflex/issues/242


storybook/shared/box.js
```diff
@@ -21,14 +21,17 @@ type BoxProps = {
...
+  margin,
   style,
   type,
   value = type,
   ...props
 }: BoxProps) {
   return (
-    <Grid margin={margin} {...props}>
+    <Grid
+      margin={typeof margin === 'undefined' ? [0, 0.5, 1] : margin}
```
obartra: flow still (v.0.54) chokes on mixed types with default params for function components. To work around it, instead of adding `$flowDisable` I just moved it to `margin`

src/utils.js
```diff
@@ -91,31 +91,71 @@ export const getSpacingClasses = memoize(
...
+ * - arrays of strings or numbers (converted to float):
+ *   - `margin={["1", 0, "0.5"]}` becomes `[1, 0, 0.5]`
+ * - space-separated or comma-separated strings
+ *   - `margin="0"` becomes `[0]`
+ *   - `margin="1 0"` becomes `[1, 0]`
+ *   - `margin="1,0"` becomes `[2, 0]`
+ * - numbers
+ *   - `margin={1}` becomes `[1]`
+ */
+function parseSpacing(spacing): number[] | void {
```
obartra: This is where the bulk of the changes are

src/utils.js
```diff
@@ -91,31 +91,71 @@ export const getSpacingClasses = memoize(
   (values, type) => type + values.toString()
 )
 
-function getCSS(prop, value, base) {
+function getCSS(
+  prop: string,
+  value: number | string,
+  base: number
+): { [string]: number } {
+  const num = typeof value === 'number' ? value : parseFloat(value || 0)
```
obartra: For consistency, also allow `marginTop` and similar properties to be used as a string. So if you can do `margin="1"` you should also be able to do `marginTop="1"` although it doesn't seem that common of a scenario

src/utils.spec.js
```diff
@@ -3,18 +3,91 @@ import { log, combineSpacing, validateSpacingProps } from './utils'
...
+  )
+  ;[
+    (
+      '1 0.5 2 0',
+      '1,0.5,2,0',
+      '1   0.5  2 0',
+      '1, 0.5, 2, 0',
+      '1 , 0.5  , 2 ,  0'
+    ),
+  ].forEach(margin =>
```
obartra: Added the same test cases than the linked regex in the comments

storybook/stories/components/card/composite.js
```diff
@@ -55,21 +55,21 @@ export default function() {
...
+                  <Grid dev={1} padding="1, 0.5, 0">
                     <Col size={6}>
                       <h3>Title</h3>
                     </Col>
                     <Col size={6} justify="right">
                       test
                     </Col>
                   </Grid>
-                  <Grid padding={[1, 0.5, 0, 0.5]}>
+                  <Grid padding="1, 0.5, 0">
```
obartra: I updated a couple examples to demonstrate how these changes don't impact snapshot tests

Issue #238, opened 2017-08-29
cropping images too early


obartra:
As this error suggests: https://circleci.com/gh/obartra/reflex/1668#tests/containers/1:

```shell
 safari    ✖ Error: There are some read requests waitng on finished stream
 safari       at ReadStream.onend (_stream_readable.js:584:10)
    at endReadableNT (_stream_readable.js:1047:12)
    at _combinedTickCallback (internal/process/next_tick.js:102:11)
    at process._tickCallback (internal/process/next_tick.js:161:9)
 safari   FAILED:  1 errors and 13 passed (43.316s)
```

Either sharp is starting to crop before the image is ready (saucelabs issue) or it reports it's done cropping when it's still using it (sharp issue). It's probably the later but we should investigate

obartra:
I haven't seen this issue since [this](https://github.com/obartra/reflex/pull/241/files#diff-2c29da9c51430f860f51bf746942e29fR28) change but I'll keep the ticket open a bit longer just to make sure

PR #245, opened 2017-09-03
fix(ie11): partial fix for ie11 example

arahansen:
this fixes an example that was incorrectly using reflex by applying flex properties inline.

This does however expose another ie11 bug: if [this line ](https://github.com/obartra/reflex/compare/arahansen/ie11-fix?#diff-5d2fd66d9ac2b850a199b52b97066957R38) were to use `size="auto"`, as it was previously, the layout will be brokene with `Lorem ipsum` text extending beyond the max length of its container.


obartra:
Good catch, this is good as is but we should also add an example of the broken IE test (we can put it in a separate PR).

With these changes, are we able to reduce the tolerance threshold? I've noticed some more false positives since I added screenshot cropping. Have you seen similar issues?

arahansen:
Haven't looked at CI. I can play with reducing tolerance. Does screenshot cropping need to be reduced?

obartra:
No, the sizing works correctly but something says it's done before the file is closed, I tried to add more details here https://github.com/obartra/reflex/issues/238

PR #263, opened 2017-09-11
test(image): Add suite retries to reduce false positives

obartra:


obartra:
Closes https://github.com/obartra/reflex/issues/238

Issue #270, opened 2017-09-22
new margin syntax bug with Col components


andrewmtoy:
`<Col margin={0}>Thing</Col>` produces this error: 
`Cannot define margin or padding and a direction at the same time`

PR #272, opened 2017-09-28
fix(error-logging): dont show error message for valid props

arahansen:
Closes https://github.com/obartra/reflex/issues/270

BREAKING CHANGE


src/derived/col.js
```diff
@@ -12,9 +11,7 @@ const defaults = {
 }
 
 export default function Col(props: Props) {
-  validateSpacingProps(props)
```
arahansen: Base.js is already calling validateSpacingProps

obartra:
@andrewmtoy this should fix the bug report you opened

PR #273, opened 2017-10-01
fix(withGrid): match export name to file name

arahansen:
- webpack `external` wasn't properly pointing to the correct dependency
- export was improperly named
- added `GridProps` type to export as a helper for types

BREAKING CHANGE

### DIFF
--- a/scripts/webpack.config.js
+++ b/scripts/webpack.config.js
@@ -68,7 +68,7 @@ module.exports = {
   devtool: 'source-map',
   externals: {
     react: 'react',
-    'prop-types': 'PropTypes',
+    'prop-types': 'prop-types',
   },
   module: {
     rules: [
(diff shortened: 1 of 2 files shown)


scripts/webpack.config.js
```diff
@@ -68,7 +68,7 @@ module.exports = {
   devtool: 'source-map',
   externals: {
     react: 'react',
-    'prop-types': 'PropTypes',
```
obartra: This is the right fix but have you verified if there are any changes in the build?

arahansen: could you elaborate on what you mean by "changes in the build"?

obartra: as in if looking at what's bundled before/after these changes shows any differences

arahansen: ah! I did! react and prop-types are properly referenced as external dependencies.

PR #281, opened 2017-11-17
add convenient spacing names

andrewmtoy:
screen shot 2017-11-20 at 11 53 09 am


src/utils.js
```diff
@@ -1,9 +1,11 @@
 // @flow
-import { memoize } from 'lodash'
-import type { SpacingProps, Noop } from './types'
+import { memoize, isString, isFinite, flow } from 'lodash'
```
obartra: All this methods will increase the bundle size, any time we can do without them in reflex we should. It's a bit of a balancing act between what's already in use and how much value new ones add

storybook/stories/spacingNames/default.js
```diff
@@ -0,0 +1,19 @@
+// @flow
```
obartra: This falls within the other usage examples I would put it in there rather than on a new section

src/utils.spec.js
```diff
@@ -189,9 +195,20 @@ describe('getCSS', () => {
...
+
+    expect(css).toEqual({
+      borderBottomWidth: spacingNames.XL,
+    })
+  })
 })
 
 describe('parseSpacing', () => {
+  it('should return undefined if spacing is undefined', () => {
+    expect(parseSpacing(undefined)).not.toBeDefined()
```
obartra: This case only happens in test as far as I can tell and adds additional logic to the code, let's remove it

src/utils.js
```diff
@@ -113,19 +129,48 @@ export function getCSS(
  * - numbers
  *   - `margin={1}` becomes `[1]`
  */
-export function parseSpacing(spacing: any): number[] | void {
-  if (spacing instanceof Array) {
-    return spacing.map(parseFloat)
-  } else if (typeof spacing === 'undefined') {
+
+export function parseSpacing(spacing: any): Array<number> | void {
```
obartra: This method is now doing a lot. What if, before getting here we just replace all aliases with their values? That way we have to do a lot less changes here

This we are also loosing the `base` meaning which we should keep. alias * base should still work

src/defaults.json
```diff
@@ -0,0 +1,15 @@
...
+    "S/2": "4px",
+    "S": "8px",
+    "M/2": "8px",
+    "M": "16px",
+    "L/2": "12px",
+    "L": "24px",
+    "XL/2": "16px",
+    "XL":"32px",
+    "XXL/2": "24px",
+    "XXL":"48px"
```
obartra: These values should be multiples of base (see comment below), that way we can just say `"XS": 0.5` and so on

andrewmtoy: I thought the idea was to eventually get rid of base entirely and switch over to using only aliased spacing names.  Is that not the case?

andrewmtoy: So one disadvantage of doing it this way is that we'd have to write base={4} for every component that uses aliases until we completely switch over to this api. 

```
<Col base={4} margin={['XS', 'S/2']} padding="S,M/2">
  My paddings and margins are set using spacingNames!
</Col>
```

andrewmtoy: I could also pass base through the provider.

obartra: You don't have to pass base to col, `24` the same way you don't pass the multiplier. You can rely on the default value. I agree that we can eventually set it through provider but you can also write a wrapper component that changes the defaults it if needed. `XS` should really just be an alias for the number you would type otherwise, in this case `0.5`

storybook/stories/spacingNames/default.js
andrewmtoy: I think having a separate story with notes would be really helpful.  I'll put it under the Grid folder.  This is related to both the margin and spacing story but I don't want to put it in both or just one of them.

src/utils.spec.js
andrewmtoy: This is actually valid.  This is the correct default behavior when a Grid/Col component is called with no margin or padding properties.  I'm going to leave this test here.

src/utils.js
```diff
@@ -113,31 +113,70 @@ export function getCSS(
...
+export function replaceAliases(
+  spacingArray: Array<SpacingValues>,
+  spacingAliases?: SpacingAliases
+): Array<SpacingValues> {
+  if (!spacingAliases) {
+    return spacingArray
+  }
+  return spacingArray.map(value => {
+    if (
+      spacingAliases &&
```
andrewmtoy: running into flow issues here.

src/utils.spec.js
```diff
@@ -215,4 +222,49 @@ describe('parseSpacing', () => {
 
     expect(log.error).toHaveBeenCalled()
   })
+
+  it('should accept strings and numbers in an array, space separated strings, and comma separated strings', () => {
```
andrewmtoy: added some missing tests

src/spacingAliasesProvider.js
```diff
@@ -0,0 +1,26 @@
...
+}
+
+type Props = {
+  spacingAliases: SpacingAliases,
+  children: React$Node,
+}
+
+export const SpacingAliasesContext = { spacingAliases: PropTypes.object }
+
+export default class SpacingAliasesProvider extends React.Component<Props> {
```
arahansen: Should we just call this `Provider` (and export it as such) and use it as a single location for all configuration moving forward?

src/spacingAliasesProvider.js
```diff
@@ -0,0 +1,26 @@
...
+export type SpacingAliases = {
+  [spacingAlias: string]: number,
+}
+
+type Props = {
+  spacingAliases: SpacingAliases,
+  children: React$Node,
+}
+
+export const SpacingAliasesContext = { spacingAliases: PropTypes.object }
```
arahansen: I think we should namespace this to something like `xnReflexSpacingAliases`, or `xnReflex: { spacingAliases }` since this is a global namespace we are tapping into.

obartra: This shouldn't be specific to spacing though the idea is that we'll use the provider for any other variables (number of columns, breakpoints...)

src/utils.spec.js
obartra: You are right, this test makes sense

src/spacingAliasesProvider.js
andrewmtoy: I don't think this should be prefixed with xn.  What about renaming to ConfigProvider?

src/spacingAliasesProvider.js
andrewmtoy: What do you think about ConfigProvider?

src/spacingAliasesProvider.js
arahansen: I want something specific to the library name, `xn-reflex`

andrewmtoy: Oh I see what you're saying.

obartra:
@andrewmtoy you need to accept the contributor request for CI to run

andrewmtoy:
Cool, just accepted.  Can you trigger a rebuild for me please?

PR #283, opened 2017-11-21
feat(alias): Add alias support for spacing values

andrewmtoy:


obartra:
we'll need to document these changes on the wiki but we can do this afterwards

PR #285, opened 2017-11-24
feat(provider): Add base to ConfigProvider

obartra:
- Move default values to `defaults.json`
- Add `base` for `Grid` as a configurable option


obartra:
@arahansen  I forgot why we implemented `Grid` and `withGrid` separately, wouldn't `Grid` just be `withGrid('div')`? Not sure if there were some implications for `ref`? Same thing with `Base` and `withBase`

arahansen:
I think Grid could definitely be `withGrid('div')`. I think grid.js was implemented before we realized we could use strings as arguments to `withGrid`

Issue #251, opened 2017-09-06
Support responsive layout


obartra:
We want to support 3 layouts:
- `small`: 0- 599
- `medium`: 600-899
- `large`: 900 +

### Rationale

Adding support for only 3 media queries will have a small footprint and allow us to enforce these 3 breakpoints. If that's not desired behavior one can always use [react-responsive](https://github.com/contra/react-responsive).

This is a commonly supported feature in a grid system and we should be able to handle it without dependencies.

### Differences

The main difference from other grid systems is that instead of attempting to reposition elements automatically, we simply hide or show them. This is intentional since that approach only works for a limited set of cases and this one gives you greater flexibility. For instance:

```js
<Grid show="small">
  <ComponentA size={12} />
  <ComponentB size={12} />
</Grid>
<Grid show="medium,large">
  <ComponentA size={6} />
  <ComponentB size={6} />
</Grid>
```

`small` layout stacks components, `medium` or `large` places them side-by-side. This is no different than other systems but now have the flexibility to decide how things are rendered. E.g.: `ComponentB` could be omitted in `mobile` or replaced by `ComponentC` and you are able to set different properties to them.

### Documentation

- These changes should include at least 1 new usage example in Storybook
- Update wiki article on [Responsive Grids](https://github.com/obartra/reflex/wiki/Responsive-Grids)

### Testing

- Add unit tests for new feature
- Add different resolution screenshots to ensure responsive layout is working

### API

The proposed API would add a `show` property to `<Base />` and `<Layout />` elements that will default to always show:

```js
<Layout show={['small', 'medium', 'large']} />
<Grid show={['small', 'medium', 'large']} />
```

Values will be customized much like `margin`. This means that values like `show="medium"` would also work (and just show on `medium` dimensions) and `show="medium, large"` or `show={['medium', 'large']}` would only exclude small screens.

PR #289, opened 2017-11-25
feat(layout): Add support for responsive layout

obartra:
- Use provider to configure breakpoints through `displayAliases`

Closes https://github.com/obartra/reflex/issues/251


src/withResolution.js
```diff
@@ -0,0 +1,86 @@
...
+      super(props)
+
+      this.state = {
+        shouldShow: true,
+      }
+      this.addMediaQueryListener(props.show)
+    }
+
+    componentWillReceiveProps({ show }: Props) {
+      this.addMediaQueryListener(show)
```
arahansen: this will remove and re-add the listener on every render?

src/withResolution.js
```diff
@@ -0,0 +1,86 @@
...
+import { ConfigContextPropTypes } from './configProvider'
+
+type Props = {
+  show?: DisplayValues,
+  base?: number,
+}
+
+type State = { shouldShow: boolean }
+
+export default function withResolution(Component: *) {
```
arahansen: Just a heads up: the usage of web-specific APIs will prove problematic for React Native (`window`,

obartra: do we have a way to detect screen resolution changes that would work well with RN?

src/withResolution.js
obartra: Good call, I can make it run only when `show` changes

src/withResolution.js
arahansen: There are built in APIs to RN that detect resolution. They're not compatible with web APIs though. 

Probably fine to go this route for now. And punt on resolution detection, since we don't even have a native solution in place 😛 .

I think the use case for a `show` api is rarer for native. Maybe something more focused on changing `size` attributes based on screen size. the `show` api can accomplish that, but it'll be a bit verbose.

obartra: Ok for now I added a step to skip `withResolution` if `window.matchMedia` is not available and log a warning. As you suggest, we can worry about a proper solution when we actually support RN.

I imagine we may want different layouts in landscape vs portrait mode or something like that but probably not much more beyond that

obartra:
94bb07e changes from tracking whether to mount/unmount the hidden components to setting `display: none` when it needs to be hidden.

obartra:
Changed approach based on discussion with @arahansen, now I added a `mediaQuery.js` to manage it. This ensures there are at most as many eventListeners are `displayAliases` keys.

It now generates an object in `withResolution` that looks like `{ small: true, large: false}` so that `Base` *could* consume it easily if we wanted to change the API for `margin` / `padding` / `size` to support a format like: `size={{ small: 6, large: 12 }}`

package.json
```diff
@@ -114,7 +114,7 @@
   "bundlesize": [
     {
       "path": "./dist/reflex.min.js",
-      "maxSize": "8 Kb"
+      "maxSize": "9.5 Kb"
```
obartra: This feature is fairly large as it is now (1.5kb more), I'm open to removing features if we think it's worth it

src/reflex.js
```diff
@@ -1,30 +1,38 @@
...
-  Overflow,
+  GridProps,
   Height,
-  Dev,
+  Justify,
+  Overflow,
+  Size,
   SpacingProps,
+  SpacingValues,
 } from './types'
```
obartra: I grouped the exports and moved `GridProps` to types now that it's gotten a bit more complex

src/types.js
```diff
@@ -59,9 +63,70 @@ export type SpacingAliases = {
...
+  marginTop?: SpacingValues,
+  padding?: Spacing,
+  paddingBottom?: SpacingValues,
+  paddingLeft?: SpacingValues,
+  paddingRight?: SpacingValues,
+  paddingTop?: SpacingValues,
+  show?: DisplayValues,
+  size?: Size,
+  style?: { [string]: string | number },
+}
```
obartra: `OneResolutionGrid` is used internally only, after `withResolution` has resolved the values to use

src/types.js
```diff
@@ -59,9 +63,70 @@ export type SpacingAliases = {
...
+  'marginLeft',
+  'marginRight',
+  'marginTop',
+  'padding',
+  'paddingBottom',
+  'paddingLeft',
+  'paddingRight',
+  'paddingTop',
+  'size',
+]
```
obartra: These are the properties for which you can now pass either an object (with keys indicating when to apply it) or the previous format

src/utils.js
```diff
@@ -40,6 +44,42 @@ export function validateSpacingProps(props: SpacingProps) {
...
+  const response = []
+  const [from, to] = displayAliases[range].split('-').map(toInt)
+
+  if (from) {
+    response.push(`(min-width: ${from}px)`)
+  }
+  if (to) {
+    response.push(`(max-width: ${to}px)`)
+  }
+  return response.join(' and ')
```
obartra: This is one of the limitations of this approach. Right now the only media queries handled are `min/max width`)

src/withResolution/mediaQuery.js
```diff
@@ -0,0 +1,63 @@
...
+    +mediaQuery: MediaQueryList,
+    +allCallbacks: (mql: MediaQueryListEvent) => void,
+  |},
+}
+type PerCallbackData = {
+  [query: string]: Array<{|
+    +alias: string,
+    +callback: Callback,
+  |}>,
+}
```
obartra: media queries manages setting up the minimal number of listeners possible and removing them when they are no longer in use, based on some quick jsfiddles this roughly doubles performance on Chrome

storybook/stories/grid/resolution.js
```diff
@@ -0,0 +1,68 @@
...
+    >
+      <Grid size="auto">Visible on small and medium screens (&lt; 900px)</Grid>
+      <Button />
+    </Grid>
+    <Grid
+      dev={2}
+      padding={2}
+      margin={[1, 0]}
+      show={['small', 'large']}
+      size={{ small: 12, large: 6 }}
```
obartra: I updated the stories to contain both the `show` property and the new `object` format (for some `size` props)

src/defaults.json
```diff
@@ -1,4 +1,9 @@
 {
   "spacingAliases": {},
+  "displayAliases": {
+    "small": "-599",
```
arahansen: It's nice how terse this syntax is, but it's a bit weird to read. Would you still prefer this to something like:
`{ small: { minWidth: 599 }, medium: { minWidth: 600, maxWidth: 899 }, ... }`

This would also be a little bit closer to how media queries are typically written (with the `minWidth/maxWidth` keywords)

src/types.js
```diff
@@ -59,9 +63,70 @@ export type SpacingAliases = {
...
+  paddingBottom?: SpacingValues,
+  paddingLeft?: SpacingValues,
+  paddingRight?: SpacingValues,
+  paddingTop?: SpacingValues,
+  show?: DisplayValues,
+  size?: Size,
+  style?: { [string]: string | number },
+}
+
+export const resolutionProperties = [
```
arahansen: This is conflating flow types, and properties used by reflex. They are different thing in my mind, and maybe shouldn't be grouped together in the same module.

src/types.js
```diff
@@ -59,9 +63,70 @@ export type SpacingAliases = {
...
+  marginTop?: SpacingValues | { [string]: SpacingValues },
+  padding?: Spacing | { [string]: Spacing },
+  paddingBottom?: SpacingValues | { [string]: SpacingValues },
+  paddingLeft?: SpacingValues | { [string]: SpacingValues },
+  paddingRight?: SpacingValues | { [string]: SpacingValues },
+  paddingTop?: SpacingValues | { [string]: SpacingValues },
+  size?: Size | { [string]: Size },
+}
+
+export default undefined
```
arahansen: I dont think you need to make eslint happy with this since you have multiple exports.

src/withResolution/mediaQuery.js
```diff
@@ -0,0 +1,63 @@
...
+
+function callCallbacks(query) {
+  return () => {
+    perCallbackData[query].forEach(({ callback, alias }) =>
+      callback(queryData[query].mediaQuery, alias)
+    )
+  }
+}
+
+export const supportsMatchMedia = !!window.matchMedia
```
arahansen: Could this be more robust? something like `window && typeof window.matchMedia === 'function'`

Mostly just thinking about the React Native use case where `window` will not be available.

Although the more I think about it, the more I think that RN will need its own implementation of `withResolution` since most of the logic happening in this module will not really work in a RN environment

src/withResolution/mediaQuery.js
```diff
@@ -0,0 +1,63 @@
...
+  [query: string]: Array<{|
+    +alias: string,
+    +callback: Callback,
+  |}>,
+}
+
+const perCallbackData: PerCallbackData = {}
+const queryData: QueryData = {}
+
+function callCallbacks(query) {
```
arahansen: Maybe this could be called something like `getCallCallbacks` since this is a HOF whose returned function is the one that calls callbacks?

src/withResolution/mediaQuery.spec.js
```diff
@@ -0,0 +1,79 @@
+import { register, unregister } from './mediaQuery'
+
+describe('Media Query Manager', () => {
```
arahansen: would we be able to do `spyOn(global.matchMedia)` (and/or `addListener`)  to validate that we are only adding as many listeners as we actually need?

src/types.js
obartra: 🙈

src/withResolution/mediaQuery.js
obartra: it's still a good point, we shouldn't assume window, will update

src/withResolution/mediaQuery.spec.js
obartra: oh interesting, will do

src/defaults.json
obartra: I guess it's a trade off between complexity of the API and flexibility but I like it because this approach enables us to add more formats later on (`e.g. { large: { orientation: 'landscape' } }`

src/types.js
obartra: that's fair, I was thinking about this, imo it made sense to put them here so that there's a single spot to look at when new properties are added. Maybe the right solution would be to generate this array but that also feels overkill

test/env.js
```diff
@@ -4,6 +4,29 @@ global.requestAnimationFrame = jest.fn().mockImplementation(() => {
...
+      let i = allListeners[media].length - 1
+
+      while (i >= 0) {
+        if (allListeners[media][i] === listener) {
+          allListeners[media].splice(i, 1)
+        }
+        i -= 1
+      }
+    },
+    onchange: () => allListeners[media].forEach(listener => listener({})),
```
andrewmtoy: nit: onChange

storybook/stories/grid/spacingAliases.md
```diff
@@ -34,5 +34,6 @@ Changing base to `24` changes the spacing values:
 | XXL/2       | 3     | 24   | 72px               |
 | XXL         | 6     | 24   | 144px              |
 
+Use the knobs to try different `base` values.
 
-Use the knobs to try different `base` values. For more information on how to use the `ConfigProvider` check the [wiki page](https://github.com/obartra/reflex/wiki/ConfigProvider).
+There are other parameters that can be customized through `ConfigProvider`. For more information check the [wiki page](https://github.com/obartra/reflex/wiki/ConfigProvider).
```
andrewmtoy: I think having a hide prop in addition to show would be helpful.

test/env.js
obartra: this is part of the MediaQueryList spec https://drafts.csswg.org/cssom-view/#dom-mediaquerylist-onchange

storybook/stories/grid/resolution.md
```diff
@@ -0,0 +1,21 @@
+`show` allows you to define when a given component is visible based on the screen width.
```
andrewmtoy: I think having a hide prop in addition to show would be helpful.

obartra: That's a good idea. I think as far as we don't allow having `show` and `hide` set at the same time there should be no issues. I'll take a look if I don't add it here I may create a ticket for a follow up PR

src/withResolution/mediaQuery.js
```diff
@@ -0,0 +1,63 @@
+// @flow
+type Callback = (mql: MediaQueryList, alias: string) => void
+type QueryData = {
+  [query: string]: {|
```
andrewmtoy: You could replace this with
```
ReadOnly<{
  mediaQuery: MediaQueryList,
  allCallbacks: (mql: MediaQueryListEvent) => void,
}>
```

src/withResolution/mediaQuery.js
```diff
@@ -0,0 +1,63 @@
...
+    +mediaQuery: MediaQueryList,
+    +allCallbacks: (mql: MediaQueryListEvent) => void,
+  |},
+}
+type PerCallbackData = {
+  [query: string]: Array<{|
```
andrewmtoy: You could replace this with
```
ReadOnlyArray<{|
       alias: string,
       callback: Callback,
    |}>,
```

src/withResolution/mediaQuery.js
andrewmtoy: What about `getCallbacksInvoker`?

src/withResolution/mediaQuery.js
obartra: the values themselves are read only once set but the array itself is not read only

src/withResolution/mediaQuery.js
obartra: same thing here

src/types.js
```diff
@@ -59,9 +63,70 @@ export type SpacingAliases = {
...
+  'paddingTop',
+  'size',
+]
+
+export type GridProps = {
+  ...OneResolutionGrid,
+  align?: AlignGrid | { [string]: AlignGrid },
+  justify?: Justify | { [string]: Justify },
+  margin?: Spacing | { [string]: Spacing },
+  marginBottom?: SpacingValues | { [string]: SpacingValues },
```
andrewmtoy: I don't think these flowtypes are correct. https://flow.org/en/docs/types/objects/#toc-objects-as-maps

`{[string]: SpacingValues}` tells me that I can pass in this to marginBottom:
```
marginBottom={foo: 1}
```

src/types.js
andrewmtoy: I also find this confusing.

src/types.js
andrewmtoy: If this is not reused, can we move it to the file where it is being used?

src/types.js
arahansen: I had noticed this too. The issue is that the possible keys are dynamic based on aliases. It's possible we can address this with generic typing. But I'm ok with a little less specific typing and addressing in a future PR.

andrewmtoy: Not sure I follow.  With aliases this could be something like `marginBottom?: string | $Keys<typeof spacingAliases>` -> `marginBottom="XL"`

obartra: With it you can now use:

```
<Grid marginBottom={{ large: 'M', medium 'S', small: 'XS' }} />
```

In addition to being able to say:

```
<Grid show="large" marginBottom="M" />
<Grid show="medium" marginBottom="S" />
<Grid show="small" marginBottom="XS" />
```

The difference is that with the first approach you maintain the same component so state is preserved. It won't be always possible to use the first approach which is why the second one is there.

The reason why the types are defined as `[string]` instead of `"large" | "medium" | "small"` is because these are just the defaults

src/withResolution/mediaQuery.spec.js
obartra: thinking about it more, we cannot attach a listener directly because `matchMedia` creates a new object to which you can attach queries. That's the part that's already tested

storybook/stories/grid/resolution.md
obartra: I've added https://github.com/obartra/reflex/issues/307

Issue #288, opened 2017-11-24
Replace `marginTypes.js` with Aliases


obartra:
Aliases now cover the same functionality, replace current usage of `marginTypes` with them

PR #300, opened 2017-12-02
feat(spacing): Add spacing defaults

obartra:
- Add default spacing values
- Switch base to 8 instead of 24
- Replace `marginTypes.js` with aliases

Closes https://github.com/obartra/reflex/issues/288
BREAKING CHANGE: Base is 8 instead of 24


src/derived/col.js
```diff
@@ -5,9 +5,9 @@ import type { Props } from '../withBase'
 
 const defaults = {
   marginTop: 0,
-  marginRight: 0.5,
-  marginBottom: 1,
-  marginLeft: 0.5,
+  marginRight: 1.5,
+  marginBottom: 3,
+  marginLeft: 1.5,
```
obartra: this looks rather arbitrary now but once we've moved to a CSS in JS and gutter size is also expressed as a configurable multiple of `base`, this can become `0 gutter_size/2 gutter_size` again

src/utils.js
```diff
@@ -187,7 +194,7 @@ export function combineSpacing({
   }
 
   const flatProps = {
-    ...props,
+    ...replaceAliasValues(props, spacingAliases),
```
obartra: This fixes a bug where aliases for individual values (e.g. `marginBottom`) were not replaced

storybook/shared/box.js
```diff
@@ -29,12 +29,12 @@ export default function Box({
 }: BoxProps) {
   return (
     <Grid
-      margin={typeof margin === 'undefined' ? [0, 0.5, 1] : margin}
+      margin={typeof margin === 'undefined' ? [0, 'L/2', 'L'] : margin}
```
obartra: From this point on it's updates to stories to use aliases instead of numerical values

obartra:
with `ConfigProvider` supporting `base` I'm wondering if we need a `<Base>` component anymore. We could make `<Grid>` take an additional, optional `base` parameter to override `ConfigProvider`'s value when needed

test/__snapshots__/Storyshoots.spec.js.snap
```diff
@@ -11255,10 +11255,10 @@ exports[`Storyshots Layout Auto 1`] = `
               className="base"
               style={
                 Object {
-                  "borderBottomWidth": 0,
+                  "borderBottomWidth": 24,
```
arahansen: This looks like the behavior changed, no?

src/utils.js
arahansen: Is there a test that covers this fix?

test/__snapshots__/Storyshoots.spec.js.snap
obartra: Oops, good catch, [this line](https://github.com/obartra/reflex/pull/300/files/4c0235c0d53e09d7fd1d87883f0ad23af6894ba8#diff-d13958befac094f9da077d89a15a314fR14) should be `"0 L/2"` instead of `"L L/2"`, I wonder why the screenshots didn't catch this, maybe it's below the fold?

src/utils.js
obartra: No, there isn't but there should be, will update

src/defaults.json
```diff
@@ -1,4 +1,16 @@
...
+    "M/2": 1,
+    "M": 2,
+    "L/2": 1.5,
+    "L": 3,
+    "XL/2": 2,
+    "XL": 4,
+    "2XL/2": 3,
+    "2XL": 6
+  },
+  "base": 8
```
andrewmtoy: Looking at this aggregated defaults.json, base is confusing.  Maybe we can rename to `baseInPixels`?

obartra: can solve this in the documentation if needed

test/__snapshots__/Storyshoots.spec.js.snap
```diff
@@ -10,19 +10,19 @@ exports[`Storyshots Components.Card Base 1`] = `
       style={Object {}}
```
andrewmtoy: Is this file named incorrectly?  Storyshoots.spec.js.snap.

obartra: this file is autogenerated

Issue #301, opened 2017-12-03
Improve error messages


obartra:
Error messages should have dedicated wiki articles with help, update current ones

PR #310, opened 2017-12-06
feat(logging): Improve error messages

obartra:
- [x] Adds better error messages and references to wiki articles that further explain them
- [x] Expand wiki pages error codes
- [x] Remove error messages (leave only error code) on production build

image

Closes https://github.com/obartra/reflex/issues/301


src/errors/INVALIDCSSPROP.txt
```diff
@@ -0,0 +1 @@
+Invalid css prop used, only "margin", "spacing" and directional variants (e.g. "marginTop") allowed
```
arahansen: whats the benefit of splitting error messages into individual .txt files instead of just a `errors.js` module that exports the strings?

obartra: there's no overhead to having it this way and you get a full document to write on. The goal is to encouarage longer error messages but yeah there are many ways in which this can be implemented. This seemed the cleanest even if not the simplest

arahansen: I'm not sure that's necessarily what I would want. Descriptive error messages that point to a wiki to read more? Totally. An entire essay in my console? Not so much.

obartra: this is not what this is. As I said, there are many ways this can be implemented. This allows you to write messages cleanly without worrying about anything else and has no overhead on the build

arahansen: I just find it unnecessary. But 🤷‍♂️

obartra: Let's talk about it more in person, I can try to shorten the longest one, the other ones are fairly short already. I still would like to keep them separate (not mixing the message with logic and easier formatting).

These are important messages (we don't throw to prevent a blank screen but the layout most likely will be messed up) and should ocur rarely. So I don't expect them taking over users consoles.

It's part of the efforts to try to make reflex more beginner friendly.

PR #312, opened 2017-12-10
refactor(style): Switch to a CSS in JS approach

obartra:
- This PR keeps parity with current behavior but:
  - Has a smaller footprint
  - Will enable further `ConfigProvider` customization
  - Will enable RN support

- Removes Color Dev Mode (keeps overlay) since it was unused and moves it to
  Reflex styleguide only

BREAKING CHANGE CSS Import no longer needed, Color Dev Mode removed

---

### Tasks

- [ ] Update readme demo
- [x] CSS Dev Mode overlay fix (disabled for small window width, master behavior was incorrect)
- [x] Validate Dev Mode tests
- [x] Validate rendering
- [x] Update readme (remove css references)
- [x] CSS Dev Mode remove color 
- [x] Fix Color Dev Prop Types
- [x] Apply single color class
- [x] Update Wiki


src/withResolution/index.js
```diff
@@ -11,7 +11,7 @@ import {
   hasTrueValues,
   type ShouldShow,
 } from './withResolution.logic'
-import errors from '../errors'
+import errors from /* preval */ '../errors'
```
arahansen: could we just be adding a `// @preval` comment to the top of errors.js?

test/Storyshoots.spec.js
```diff
@@ -1,6 +1,47 @@
 import initStoryshots from '@storybook/addon-storyshots'
 import './env'
 
+jest.mock('../src/cxs', () => i => {
```
arahansen: is it worth testing this in isolation beyond actually generating snapshots?

src/withResolution/index.js
obartra: tests where choking on it, if the file itself gets preval'ed then it also doesn't count towards test coverage which it's not that big of a deal but it's unfair because it is tested

test/Storyshoots.spec.js
obartra: There's a case to be made there for sure, honestly this feels more like a quick hack to get free css snapshot. We could also just return `"className"` and have separate css snapshots.

Since (afaik) there's no readily available solution for CXS snapshots and we would need to implement our own for something we already have here it has a very low ROI imo.

src/withResolution/index.js
obartra: hm maybe it's something webpack should do, alias `../errors` to something like `../errors/preval` (which imports with preval and exports) and then tests would remain unaffected

arahansen: Eh that seems more confusing and magical. Would we be able to disable preval for tests?

obartra: yeah it's a balancing act, with the current approach if you forget to import `./errors` without preval you'll import it twice (and increase bundle size), if we do it the other way there's some webpack magic.

I'm fine either way, I don't like either approach tbh

We can disable preval for tests _if_ nothing else is using preval

arahansen: I'm fine keeping it this way. Since bundlesize should catch mistakes like that.

PR #314, opened 2017-12-12
feat(grid): expose underlying ref for grids

arahansen:
Having access to the underlying DOM node can be useful for executing side effects as well as measurements on the DOM.


arahansen:
Can you merge this for me? Bithound is not finishing

PR #315, opened 2017-12-12
fix(flow): fix flow errors for external consumers

arahansen:
refactors defaults.json to js module because external flow configs cannot understand .json files within node_modules.

I can probably hack my way with custom flow definitions for this json file. But this seems like a more universal solution.

Also fixes some "sketchy null type" warnings


src/defaults.js
```diff
@@ -0,0 +1,30 @@
...
+      maxWidth: '899px',
+    },
+    large: {
+      minWidth: '900px',
+    },
+  },
+  base: 8,
+}
+
+export default undefined
```
obartra: but you are importing it as default in `configProvider`

src/grid/asGrid.js
```diff
@@ -52,8 +52,8 @@ export default function asGrid(Component: *) {
   ) {
     const classes = compact([
       className,
-      size && styles.col,
-      size && styles[`col-${String(size)}`],
+      size !== undefined && styles.col,
+      size !== undefined && styles[`col-${String(size)}`],
```
obartra: here we don't have the same guarantees on undefined, if we want to check for it we probably want to do `typeof size !== 'undefined'`

PR #327, opened 2017-12-15
fix(npm): Correct npm targets

obartra:
- Remove `module` since although we can offer an ES6 build, it has additional (unlisted) dependencies: `cxs`, `preval`
- Default to a new `index.js` file that decides which file to serve based on `NODE_ENV` (similar to React's [approach](https://github.com/facebook/react/blob/master/packages/react/npm/index.js))
- Preserve the `browser` section from `package.json` but make it point to the minified build by default (this is useful for tools like unpkg, that will use it to load the minified version instead when importing https://unpkg.com/xn-reflex)

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,3 +1,4 @@
+// @flow
 /* eslint-disable global-require */
 if (process.env.NODE_ENV === 'production') {
   module.exports = require('./dist/reflex.min.js')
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -0,0 +1,6 @@
+/* eslint-disable global-require */
+if (process.env.NODE_ENV === 'production') {
+  module.exports = require('./dist/reflex.min.js')
```
arahansen: Does flow understand this branching?

obartra: not sure how to test it, I assumed it wouldn't matter for flow

arahansen: Just tested your changes on Iris and I don't think this works the way we want it to.

PR #329, opened 2017-12-15
fix(flow): update dependencies

arahansen:
added cxs as a dependency, otherwise it won't be installed on user's machines and flow will complain it can't find cxs.

Also removing global `preval` so flow doesn't have to know about it.


src/withResolution/withResolution.logic.js
```diff
@@ -1,8 +1,10 @@
 // @flow
+import preval from 'preval.macro'
 import type { DisplayValues, DisplayAliases } from '../types'
 import defaults from '../defaults'
 import { splitPattern, log } from '../utils'
-import errors from /* preval */ '../errors'
+
+const errors = preval`module.exports = require('../errors')`
```
obartra: do you think we should change the default export for errors?

if we rename our current `/errors/index.js` to `/errors/errors.js` (and use that for tests), we can then create a `/errors/index.js` that looks like:

```js
// @preval

const errors = require('./errors').default

module.exports = errors
```

And we'll be able to consume it with less ceremony

Issue #332, opened 2017-12-16
Add `Reflex.setLogLevel(level)` method


obartra:
Allow customizing reflex log level.

Expose a `setLogLevel` method from `Reflex` with `'info' | 'warning' | 'error'` as valid parameters and that reduces the logging to that level or above

PR #333, opened 2017-12-17
feat(log): Allow custom log level

obartra:
- Add support for custom logger and log level
- Define `log.setLevel('info' | 'warn' | 'error')`
- Define `log.setLogger(logger)`

Closes https://github.com/obartra/reflex/issues/332


src/errors/index.js
```diff
@@ -1,42 +1,2 @@
...
-)
-addError(
-  errors,
-  'TOOMANYSPACEVAL',
-  `Invalid Spacing size, only first 4 values used. Spacing values follow CSS syntax, this means you can specify 1, 2, 3 or 4 values but not more.`
-)
-
-module.exports = errors
+// @preval
+module.exports = require('../errors')
```
obartra: I moved errors to `./errors/errors.js` and the `./errors/index.js` simply preval's it. This simplifies importing errors (no need to be concerned about `preval`, you can import with `import errors from './errors'`)

Issue #324, opened 2017-12-14
Remove `./variables.js`


obartra:
All values within `variables.js` should be moved to either `ConfigProvider` or to `dev` (if they are dev specific)

PR #337, opened 2017-12-18
refactor(dev): Switch Dev Mode to a component

obartra:
- This change is a pre-requisite for a proper implementation of
  https://github.com/obartra/reflex/issues/324
- Remove `initDevMode` and `toggleOverlay` method
- Add `<Dev />` Component
- Replace `<Dev />` in the production build with a no-op (~1kb smaller bundle!)
- React 15.5 is not supported for `<Dev />` (relies on portals)
- TODO: Update wiki to reflect these changes

BREAKING CHANGE Dev Mode is no longer accessible via Reflex API and is
implemented as a dedicated component


storybook/shared/wrapStory.js
```diff
@@ -0,0 +1,12 @@
+// @flow
+import * as React from 'react'
+import { Dev } from 'reflex'
```
obartra: It should import `Dev` directly from `reflex/src/dev/dev.js` to ensure the production build also shows the overlay

PR #354, opened 2018-01-05
fix #353

### LINKED ISSUE #353: index.js should not point to `src`
index.js should never point to the src of gymnast. Instead it should point to a babel-ified version of the src that does not use es6 imports. Otherwise, users need to do all transpilation of gymnast themselves.

Looks like we can just point to `gymnast/dist/gymnast.js`?

arahansen:


arahansen:
I recall this was originally put in place to match Fleets needs. Should this still cover their use case?

obartra:
I added this after you requested the default version not to be the dist (to use the uncompiled version). Ideally we would point to a transpiled but unbuilt version as it is now we should move all dependencies to dev dependencies

arahansen:
we still dependencies to stay as dependencies for flow types to work properly

obartra:
oh got it, I thought flow types were installed and checked for dev dependencies as well. Either way I think this is good to go, we can tackle the deps later if we need to

PR #358, opened 2018-01-09
feat(api): Expose defaults

obartra:
- Export defaults
- Replace "browser" with "unpkg" entry file in package.json to correctly
  import dev version in react while preserving unpkg compatibility.
- Update dependencies (except storybook)

Closes https://github.com/gymnastjs/gymnast/issues/357 https://github.com/gymnastjs/gymnast/issues/355 https://github.com/gymnastjs/gymnast/pull/342


circle.yml
```diff
@@ -22,6 +22,7 @@ test:
   override:
     - yarn build
     - scripts/citests.sh:
+        timeout: 1200
         parallel: true
```
arahansen: What issue were you seeing here?

obartra: I'm still debugging this, this is not the right solution though.

iphone7 tests were timing out (still are) I thought we had too many stories but looking at saucelabs it seems like it's not completing correctly and re-trying.

It may be related to the picturebook update

obartra: Looks like the issue is related to the storybook update and iphone 7. I'll look into it separately https://github.com/gymnastjs/gymnast/issues/359

PR #361, opened 2018-01-15
feat(bundle): Consolidate exported bundle format

obartra:
These changes ensure that any `Gymnast` export can also be imported as:

```js
import Grid from 'gymnast/src/grid'
```

The format is `gymnast/src/${export}` where export is the name of the export
but with the first letter always in lower case


test/e2e/gymnast.spec.js
```diff
@@ -0,0 +1,70 @@
...
+  'Layout',
+  'log',
+  'Offset',
+  'Root',
+  'utils',
+  'withResolution',
+]
+
+describe('e2e tests', () => {
+  it('should export the same values from the main export than individual files', () => {
```
arahansen: What if this were just a snapshot test?

obartra: This validates it's really the same export, with the snapshot test we would only validate that the export is of the same type. It's probably fine either way, this seemed more thorough but with the downside of having to manually enforce consistency. I added the test after that to make sure a test fails if an export is added and not included here.

Do you feel like it would be better with snapshot tests?

arahansen: my thought for a snapshot is that if we add an API a snapshot test would help remind us to add it to the test automatically, rather than needing to remember to manually update this definition

obartra: that's fair, ok I'll make the update

PR #372, opened 2018-02-16
feat(resolution): Expand media query support

obartra:
- Lower the min width to account for iPhone 5 resolution
- Add support for OR logic within displayAliases
  - This is to support applying a media query when either the screen or the device
    with are within certain values
- Allows all media queries to be used
  - This removes the need for whitelisting
  - Has the potential to introduce errors since now it's up to the user to ensure they are valid

### Tasks

- [x] Make cxs work with these changes
- [x] Add additional media query unit tests
- [x] Add story showcasing new behavior

BREAKING CHANGE layout defaults change
Closes https://github.com/gymnastjs/gymnast/issues/367
Closes https://github.com/gymnastjs/gymnast/issues/366


storybook/stories/configProvider/displayAliases.js
```diff
@@ -0,0 +1,38 @@
...
+      {
+        minWidth: '601px',
+      },
+    ],
+  }
+
+  return (
+    <ConfigProvider displayAliases={displayAliases}>
+      <Grid justify="center">
+        <Grid show="test" style={colors.colors1} margin="L" size={6}>
```
andrewmtoy: What do you think about renaming `show` to `showWhen`?

obartra: If we want to make changes to the resolution api now is the time (I think only PES is using it)

Since this is part of the existing API, if we change it I wouldn't make it as part of this PR but we can update it

PR #374, opened 2018-02-23
fix(flow): rename src/index.js to src/gymnast.js

arahansen:
Since webpack outputs a file called `gymnast.js`, the src file needs to be named the same so that when we output flow files, we get `gymnast.js` (webpack output) next to `gymnast.js.flow` (flow types generated from src).

Alternatively, we could update webpack to output a file named `index.js`.


.flowconfig
```diff
@@ -14,7 +14,7 @@
 [libs]
 
 [options]
-module.name_mapper='gymnast' -> '<PROJECT_ROOT>/src'
+module.name_mapper='gymnast' -> '<PROJECT_ROOT>/dist/gymnast'
```
arahansen: This should hopefully give us a better confidence that flow types are working correctly

PR #376, opened 2018-02-26
feat(asCore): Convert asCore into getCoreStyles helper

### LINKED ISSUE #370: Reduce React Trace
Due to core / withResolution and Grid / Col, gymnast pollutes the react component tree considerably.

This ticket is to explore alternatives

arahansen:
Helps reduce the depth of React tree rendering to partially fix #370. Also, some preliminary work that should make the react-native port easier.


test/__snapshots__/storyshots.spec.js.snap
```diff
@@ -9870,10 +9870,10 @@ exports[`Storyshots Config Provider Spacing Aliases 1`] = `
         style={
           Object {
             "backgroundColor": "#efefef",
-            "borderBottomWidth": 48,
```
arahansen: I'm stumped why these values changed. When loading storybook in browser, the styles are correct, and screenshot comparisons are passing. Is this a limitation with how context works within storyshots perhaps?

obartra: Oh interesting it's only affecting those set by context. It does sound like an enzyme limitation not liking the changes but not sure if there's much we can do about it.

As far as we have this covered with other tests (like the image comparison ones you mentioned) I think we are good

src/asGrid/index.js
```diff
@@ -8,17 +8,26 @@ import type {
 } from '../types'
 import { styles, getCol } from './grid.styles'
 import { getValue } from '../utils'
-import asCore from '../core/asCore'
+import { sharedResolutionProperties, getCoreStyles } from '../core'
```
obartra: there should be only a single default export from `core`. The goal is that you can either `import { x } from 'gymnast` or `import x from 'gymnast/x'` and they are both equivalent.

This allows a consumer to potentially use the [babel lodash plugin](https://github.com/lodash/babel-plugin-lodash) to reduce the size of gymnast even further.

If these values are meant to be exported they should be available as `gymnast/getCoreStyles` and if not they should live elsewhere (e.g. `gymnast/core/utils`).

Because `core` is not exported, this wasn't caught by [these tests](https://github.com/gymnastjs/gymnast/blob/master/test/e2e/gymnast.spec.js#L16) but it may be useful to keep the file structure consistent

src/asGrid/index.js
```diff
@@ -27,8 +36,11 @@ export default function asGrid(
       justify && styles[`${justify}Justify`],
     ])
 
-    return <Component {...props} ref={innerRef} className={classes.join(' ')} />
+    return <Component ref={innerRef} {...props} className={classes.join(' ')} />
```
obartra: just curious on whether this change matters for flow or something like that

src/asGrid/index.js
```diff
@@ -27,8 +36,11 @@ export default function asGrid(
...
   }
 
-  return asCore(Grid, resolutionProperties)
+  return withResolution(
+    Grid,
+    sharedResolutionProperties.concat(resolutionProperties)
```
obartra: why not keep this responsibility within `withResolution`?

src/asLayout/index.js
```diff
@@ -37,5 +39,9 @@ export default function asLayout(
 
     return <Component ref={innerRef} {...props} className={classes.join(' ')} />
   }
-  return asCore(Layout, resolutionProperties)
+
+  return withResolution(
+    Layout,
+    sharedResolutionProperties.concat(resolutionProperties)
```
obartra: same here

src/core/index.js
```diff
@@ -0,0 +1,58 @@
+// @flow
+import { combineSpacing, getValue } from '../utils'
+import type { ConfigProviderContext, OneResolution } from '../types'
+
+export const sharedResolutionProperties = [
```
obartra: I wouldn't export these and just move them as part of `getCoreStyles`

PR #383, opened 2018-08-13
feat(pageWidth): 'none' option for maxPageWidth

jeffwklein:


obartra:
These changes look good. Unfortunately something broke with picturebook. Still trying to figure out what's going on. Will commit to this PR with an updated fix. Sorry for the inconvenience

PR #390, opened 2018-08-21
chore(npm): Update storybook

obartra:
Update picturebook and screenshots


package.json
```diff
@@ -83,12 +84,12 @@
...
     "react-test-renderer": "16.4.2",
     "regenerator-runtime": "0.12.1",
-    "saucelabs": "1.5.0",
-    "selenium-standalone": "6.15.1",
+    "require-context": "1.1.0",
     "semantic-release": "15.9.9",
     "storybook-addon-jsx": "5.4.0",
     "style-loader": "0.22.1",
     "stylelint": "9.5.0",
+    "tape": "4.9.1",
```
arahansen: Why have two testing frameworks?

nightwatch.config.js
```diff
@@ -0,0 +1,34 @@
+require('babel-register')
```
arahansen: How much of this is specific to how circle CI needs to be setup?

package.json
obartra: I was lazy and just copy pasted 😆 I'll update to jest

nightwatch.config.js
obartra: it's weird because gymnast doesn't run well locally. You first need to deploy to CI and then you can run with `URL=whereveritwasdeployed yarn test:image`.

That being said, if you specify URL this also works locally (assuming storybook was deployed elsewhere)

nightwatch.config.js
```diff
@@ -0,0 +1,48 @@
+/* eslint-disable no-console */
```
arahansen: I think some comments in here describing what everything is doing could be helpful. Not being super familiar with CircleCI, some of this looks like magical incantation voodoo.

test/img.spec.js
```diff
@@ -0,0 +1,58 @@
+// @flow
```
arahansen: is this a helper test that picturebook could provide as a default?

arahansen:
Were you able to run CI a couple times to verify that all tests pass consistently?

obartra:
Yeah screenshot tests have been passing consistently locally and on CI, the previous failures on CI were not screenshot related

test/img.spec.js
obartra: it's fairly project-specific. I don't think it's really needed either, it's just a glorified version of

```
 cat ./picturebook-results.json | jq '.status' | grep "SUCCESS" > /dev/null && echo "success" || echo "failed" && exit 1
```

arahansen: lol, well that is a pretty specific command 😅 would be nice if there was a simpler way to do that.

obartra: something like `picturebook check --output ./picturebook-results.json`?

arahansen: I do like that better! would we be able to kill the jest image test in that case? or does it still provide value above anything `picturebook check` would do?

obartra: this gives us more info. It's up to us how much time do we want to invest on building a CLI tool for picturebook.

I can see some value to also use it as an interactive snapshot update and to generate the initial set of browser validations. Beyond that not sure what other use cases it could have

arahansen: 👍

obartra:
These changes highlight an issue with `fit.desktop.ie.png`

PR #391, opened 2018-08-28
refactor(context): Update to latest React Context API

derrickhnguyen:
## Description

The React team will not support the legacy context API in React 17. To keep up-to-date with latest changes and improvements from the React team, we must transition all usage of the legacy context API with the newest one.

## How Has This Been Tested?
- [x] Unit tests


src/defaults.js
```diff
@@ -60,4 +59,4 @@ export default {
     '2XL': 6,
   },
   verticalGutter: 3, // value (in base units) that separates columns vertically
-}
+}: Object)
```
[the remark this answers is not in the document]
obartra: Have you tried with `$Shape<ConfigContextType>` and `{ ...ConfigContextType }`?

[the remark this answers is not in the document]
obartra: @arahansen do you have other suggestions?

scripts/webpack.config.js
```diff
@@ -46,8 +46,6 @@ module.exports = {
   devtool: 'source-map',
   externals: {
     react: 'react',
-    'react-dom': 'react-dom',
```
obartra: is `react-dom` still needed here then?

obartra:
This is a breaking change, make sure the combined commit says "BREAKING CHANGE" on the last line following conventional changelog format (otherwise there won't be a major version increase)

scripts/webpack.config.js
derrickhnguyen: You're right, I'll add this back.

src/configContext/index.js
```diff
@@ -0,0 +1,9 @@
+// @flow
+
+import ConfigProvider from './configProvider'
+import ConfigConsumer from './configConsumer'
+
+export default {
+  Provider: ConfigProvider,
+  Consumer: ConfigConsumer,
+}
```
obartra: why export both? We only need `ConfigConsumer` externally (the provider is only used internally, I would not export it)

src/configContext/index.spec.js
```diff
@@ -0,0 +1,141 @@
...
+    wrapper = mount(
+      <ConfigContext.Provider columns={2} base={4}>
+        <ConfigContext.Provider columns={3} base={5}>
+          <TesterComponent>{render1}</TesterComponent>
+        </ConfigContext.Provider>
+        <ConfigContext.Provider base={1}>
+          <TesterComponent>{render2}</TesterComponent>
+        </ConfigContext.Provider>
+        <TesterComponent>{render3}</TesterComponent>
+      </ConfigContext.Provider>
```
obartra: 👍 nice test cases

src/utils/index.js
```diff
@@ -244,15 +243,13 @@ export function toCXS<A>(raw: {
...
+  return ([override, context[property], defaults[property]].find(
+    isDefined
+  ): any)
 }
 
 export function getValues(context: *, overrides?: * = {}) {
-  const contextValues = get(context, 'gymnast', {})
-
-  return { ...defaults, ...contextValues, ...overrides }
+  return { ...defaults, ...context, ...overrides }
```
obartra: this method looks so nice now ✨

src/configContext/index.js
derrickhnguyen: Make sense. I'll make Provider only exportable inside it's own file but not in `gymnast.js`

src/defaults.js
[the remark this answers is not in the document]
arahansen: this worked for me:
```
{|
  ...ConfigContextType,
  gutter: number,
  fallbackDisplayKey: string,
  verticalGutter: number,
|}
```

src/configContext/index.js
derrickhnguyen: As I was working on removing this, I noticed (via Storybook) that when using `<ConfigProvider>` imported from a relative path alongside a component (i.e. `<Grid>`) imported from a `gymnast` import, that `<Grid>` will not use the same `React.createContext()` as provided from `<ConfigProvider>` when imported from a relative path.

This is only an issue in Storybook in the `Config Provider` section. However, this shouldn't be a problem in production, since we're only exporting `Consumer`, which should be using the same `React.createcontext()` as the `Provider` imported from `gymnast`.

Example:
```
// @flow
// storybook/stories/configProvider/columns.js

import * as React from 'react'
import { times } from 'lodash'
import { number } from '@storybook/addon-knobs'
import { Grid } from 'gymnast'
import { colors } from '../../shared'

// This is using a different `React.createContext()` than the imported
// Grid above. Therefore, Grid isn't going to consume the 'columns' that
// was passed through from this ConfigProvider, and instead, will used
// the default columns value.
import ConfigProvider from '../../../src/configContext/configProvider'

export default () => {
  const columns = number('Columns', 10, { range: true, min: 1, max: 24 })
  const items = number('Items', 20, { range: true, min: 1, max: 48 })

  return (
    <ConfigProvider columns={columns}>
      <Grid padding="0 L/2">
        {times(items, key => (
          <Grid key={key} padding="0 L/2 L" size={1}>
            <Grid
              style={colors.colors1}
              padding="L/2"
              align="center"
              justify="center"
            >
              {key + 1}
            </Grid>
          </Grid>
        ))}
      </Grid>
    </ConfigProvider>
  )
}
```

This appears to be a known issue via this [issue](https://github.com/facebook/react/issues/13346)

To alleviate the Storybook problem, I can remove all imports from `gymnast` and just imported via relative path, like so:

```
// @flow
// storybook/stories/configProvider/columns.js

import * as React from 'react'
import { times } from 'lodash'
import { number } from '@storybook/addon-knobs'
import { colors } from '../../shared'
import ConsumerProvider from '../../../src/configContext/configProvider'
import Grid from '../../../src/grid'

export default () => {
  const columns = number('Columns', 10, { range: true, min: 1, max: 24 })
  const items = number('Items', 20, { range: true, min: 1, max: 48 })

  return (
    <ConsumerProvider columns={columns}>
      <Grid padding="0 L/2">
        {times(items, key => (
          <Grid key={key} padding="0 L/2 L" size={1}>
            <Grid
              style={colors.colors1}
              padding="L/2"
              align="center"
              justify="center"
            >
              {key + 1}
            </Grid>
          </Grid>
        ))}
      </Grid>
    </ConsumerProvider>
  )
}
```

obartra: We could add a babel alias to it to make it a bit more convenient to use. If we define one (for stories only) that maps `gymnastSrc` to the `src` path we could then do something like `gymnastSrc/configContext/configProvider`

test/e2e/__snapshots__/gymnast.spec.js.snap
```diff
@@ -3,7 +3,10 @@
 exports[`e2e tests should export the same values from the main export than individual files 1`] = `
 Object {
   "Col": [Function],
-  "ConfigProvider": [Function],
```
obartra: with the update to not export config context, will we keep the same API?

src/gymnast.js
```diff
@@ -3,7 +3,7 @@ import './cxs'
 
 // Components
 export { default as Col } from './col'
-export { default as ConfigProvider } from './configProvider'
+export { default as ConfigContext } from './configContext'
 export { default as Grid } from './grid'
 export { default as Layout } from './layout'
 export { default as Offset } from './offset'
```
obartra: should we export `withContext` for consistency as well?

test/e2e/__snapshots__/gymnast.spec.js.snap
derrickhnguyen: Once I make my changes, it will look like:
```
Object {
  "Col": [Function],
  "ConfigContext": Object {
    "Consumer": [Function],
  },
```

or another option to create a separate folder for `configConsumer`, then the snapshot could look like:
```
Object {
  "Col": [Function],
  "ConfigConsumer": [Function],
```

src/configContext/index.js
obartra: to ensure the project is still tree-shakeable with [babel-plugin-lodash](https://github.com/lodash/babel-plugin-lodash) this `index.js` file needs to contain as default export the same value we export through `src/index.js`

test/e2e/__snapshots__/gymnast.spec.js.snap
obartra: why not keep the API the same? if we change it we'll have to have a separate v1 vs v0 docs

obartra: Internally you see it as Provider / Consumer because we are implementing it using React Context API but the gymnast consumer should not be aware of these details.

Here we are exposing a ConfigProvider for the user to customize gymnast (a Component they can use to _provide_ configuration options).

It's a common pattern (both in naming and behavior) that [Redux](https://github.com/reduxjs/react-redux/blob/master/docs/api.md#provider-store) and [Apollo](https://www.apollographql.com/docs/react/essentials/get-started.html#creating-provider) use.

derrickhnguyen: Okay, the way Redux and Apollo are exposing their context is by externally exporting their `Provider` component while leaving their `Consumer` component un-exported, but providing a HOC component to access the context value (i.e. connect() in Redux and `<Query>` in Apollo).

So should your comment via `src/configContext/index.js` be reversed where we should externally export our `Provider` and not our `Consumer`?

obartra: whichever one we export here should be the one exported as default on `configContext/index.js` or the lodash plugin will export the wrong thing

src/configProvider/index.js
```diff
@@ -1,65 +1,42 @@
...
-        gutter,
-        verticalGutter,
-        ...props,
-      },
+      ...state,
+      ...restProps,
     }
   }
+
+  state = {}
```
derrickhnguyen: Decided to keep this here, since getting warning via `yarn test`:

```
$ jest -u
 PASS  src/asCol/asCol.spec.js
 PASS  src/asLayout/layout.spec.js
 PASS  src/offset/offset.spec.js
 PASS  src/root/root.spec.js
 PASS  src/dev/dev.spec.js
 PASS  src/withResolution/withResolution.logic.spec.js
 PASS  src/core/core.spec.js
 PASS  src/withResolution/withResolution.spec.js
 PASS  src/log/log.spec.js
 PASS  src/withResolution/mediaQuery.spec.js
 PASS  src/errors/index.spec.js
 PASS  src/utils/utils.spec.js
 PASS  src/asGrid/grid.spec.js
 PASS  src/configProvider/index.spec.js
  ● Console

    console.error node_modules/fbjs/lib/warning.js:33
      Warning: ConfigProvider: Did not properly initialize state during construction. Expected state to be an object, but it was undefined.

 PASS  test/e2e/gymnast.spec.js
 PASS  src/withContext/index.spec.js
  ● Console

    console.error node_modules/fbjs/lib/warning.js:33
      Warning: ConfigProvider: Did not properly initialize state during construction. Expected state to be an object, but it was undefined.


Test Suites: 16 passed, 16 total
Tests:       3 skipped, 118 passed, 121 total
Snapshots:   4 passed, 4 total
Time:        4.683s
Ran all test suites.
✨  Done in 6.51s.
```

PR #394, opened 2018-09-13
refactor(context): fix context performance

derrickhnguyen:
As discovered by @arahansen, in his own words:

> Because React's new context API relies on object reference to determine when to re-render children, [gymnast context] would re-render on every render cycle because we were defining a new object value={{ ...context, ...this.state}}. This change moves the object spread outside of the component so that it can be accessed within the component lifecycle hooks.

[React caveats](https://reactjs.org/docs/context.html#caveats)


derrickhnguyen:
I have no idea what you're talking about @arahansen :wink:

PR #403, opened 2018-12-02
test(enzyme): Switch to react-testing-library

obartra:
Switching to react testing library instead of enzyme.

This change was prompted because when working on the hooks branch tests wouldn't pass until enzyme supported them.


src/dev/dev.spec.js
```diff
@@ -1,69 +1,63 @@
...
-
-  event.keyCode = key
-  event.ctrlKey = ctrl
-  event.shiftKey = shift
-  event.metaKey = meta
-
-  event.initEvent('keydown', true, true)
-  document.body.dispatchEvent(event)
+function keyPress(baseElement, key, { ctrl, shift, meta }) {
+  fireEvent.keyDown(baseElement, {
```
arahansen: this is so much cleaner

PR #404, opened 2018-12-10
refactor(dev): Remove ReactDOM dependency

obartra:
- ReactDOM was only used by the `<Dev/>` component
- It was not necessary to position the element since it already uses position fixed
- This changes have no impact on the bundle size because the minified version of gymnast already noop'ed `<Dev />`

BREAKING CHANGE: Portal no longer used to position `<Dev />` component


src/dev/dev.js
```diff
@@ -26,6 +19,8 @@ type State = {|
   showOverlay: boolean,
 |}
 
+const body = (document.body: any)
```
arahansen: flow makes this optional in part to support server side rendering.

obartra: yeah, we were not handling it gracefully before either (
https://github.com/gymnastjs/gymnast/pull/404/files#diff-6cfc93b07368555038269f630aa16f54L8) but no reason not to fix it, will update

